PR #8, opened 2021-11-16
Banner redesign

benjag:
Vue 3 version of Cedar 11's banner redesign.

- Uses new `<script setup>` syntax.
- Noticed some weirdness with using icon comps in the unit tests, replaced with dummy `<div>`'s for now.


src/components/banner/CdrBanner.vue
```diff
@@ -1,36 +1,69 @@
...
-  },
-  setup(props) {
-    const baseClass = 'cdr-banner';
-    const typeClass = computed(() => props.type && buildClass(baseClass, props.type));
-    return {
-      style: useCssModule(),
-      baseClass,
-      typeClass,
-    };
+const ctx = useSlots();
```
KenjiCrosland, line 45: If I'm reading this right, the `useSlots()` method here returns an array of slotted elements? So when we define the constants `hasIconLeft`, `hasIconRight`, etc these refer to the elements themselves?

benjag: It's a helper runtime function that returns the slots object (i.e. what slots the component is using at runtime). It's essentially the equivalent of this using JSX syntax https://github.com/rei/rei-cedar/blob/next/src/components/banner/CdrBanner.jsx#L50

KenjiCrosland: Got it. That makes sense

PR #13, opened 2021-12-13
Text setup

benjag:
- Text component now uses `script setup` sugar
- Brings heading display examples over from main library

### DIFF
--- a/src/components/text/CdrText.vue
+++ b/src/components/text/CdrText.vue
@@ -1,29 +1,22 @@
 <template>
   <component
-    :is="tag"
+    :is="props.tag"
     :class="style['cdr-text']"
   >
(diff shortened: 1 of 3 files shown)


src/components/text/CdrText.vue
```diff
@@ -1,29 +1,22 @@
...
+    :is="props.tag"
     :class="style['cdr-text']"
   >
     <slot />
   </component>
 </template>
 
-<script>
-import { useCssModule, defineComponent } from 'vue';
+<script setup>
```
mhewson, line 10: nice that you don't need the return now

mhewson (approved):

KenjiCrosland (approved):

PR #16, opened 2021-12-23
Button refactor

benjag:
- [x] Script setup
- [x] Tests pass
- [x] BuildClass removed
- [x] Remove elevated? Yes, design sign off

### DIFF
--- a/src/components/button/CdrButton.vue
+++ b/src/components/button/CdrButton.vue
@@ -62,6 +62,7 @@ const baseClass = 'cdr-button';
 const style = useCssModule();
 
 /************************  Computed properties *************************/
+const buttonType = computed(() => props.tag === 'button' ? props.type : null);
 const modifierClass = computed(() => `${baseClass}--${props.modifier}`);
 const elevatedClass = computed(() => props.elevated && `${baseClass}--elevated`);
 const fullWidthClass = computed(() => !props.iconOnly && props.fullWidth
--- a/src/components/button/CdrButton.vue
+++ b/src/components/button/CdrButton.vue
@@ -1,3 +1,83 @@
+<script setup>
+import { useCssModule, computed, useSlots } from 'vue';
+
+import mapClasses from '../../utils/mapClasses';
+import { responsiveModifyClass, buildBooleanClass } from '../../utils/buildClass';
+import propValidator from '../../utils/propValidator';
+
+const slots = useSlots();
+const props = defineProps({
+tag: {
+  type: String,
+  default: 'button',
+  validator: (value) => propValidator(value, ['button', 'a']),
+},
+type: {
+  type: String,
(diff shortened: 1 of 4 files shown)


src/components/button/examples/demo/CTA.vue
```diff
@@ -14,7 +14,6 @@
         :key="index2"
         :size="button.size"
         :full-width="button.fullWidth"
-        :disabled="button.disabled"
```
benjag, line 17: This is not a valid attribute value (nor is it needed for this example). I think the JSX compilation must have ignored it

src/components/button/styles/CdrButton.module.scss
```diff
@@ -119,7 +119,7 @@
       @include cdr-button-large-mixin;
     }
 
-    &--fullWidth\@xs {
+    &--full-width\@xs {
```
benjag, line 122: For some reason, this was camelCased here where every other style modifier is kebab-cased. Changing for consistency plus the JS generates a kebab'd class ... that was a fun bug to chase down.

src/components/button/styles/vars/CdrButton.vars.scss
```diff
@@ -389,7 +380,7 @@
 }
 
 @mixin cdr-button-large-mixin() {
-  @include cdr-button-base-text-mixin-large;
+  @include cdr-button-base-text-mixin-medium;
```
benjag, line 383: This is just matching a change we made to prod

src/components/button/CdrButton.vue
```diff
@@ -21,105 +97,5 @@
...
-<!-- NOTE: REMOVED icon slot. use icon-left -->
-<script>
-import { defineComponent, useCssModule, computed } from 'vue';
-
-import mapClasses from '../../utils/mapClasses';
-import { buildClass, buildBooleanClass } from '../../utils/buildClass';
-import propValidator from '../../utils/propValidator';
-
-export default defineComponent({
-  name: 'CdrButton',
```
KenjiCrosland, line 33: Interesting. Now that we're not providing a name property are we getting it from the filename?

src/components/button/CdrButton.vue
```diff
@@ -1,3 +1,79 @@
+<script setup>
```
KenjiCrosland, line 1: Nice. The `<script setup>` syntax is a lot easier to parse and remember what goes where. It's definitely an improvement over what we're replacing.

KenjiCrosland (approved):

src/components/button/CdrButton.vue
benjag: No, `<script setup>` doesn't allow the name prop. If for some reason we end up needing it, you can declare it within a normal script block.

PR #18, opened 2021-12-28
Refactor cdr-card to use <script setup> syntax

KenjiCrosland:


src/components/card/CdrCard.vue
```diff
@@ -1,3 +1,17 @@
+<script setup>
+import { useCssModule, defineProps } from 'vue';
```
benjag, line 2: You don't need to import `defineProps` it's a compiler macro that's built in to the syntax.

benjag (approved):

PR #21, opened 2022-01-21
refactor(chip): Use script setup

benjag:


src/components/chip/CdrChip.vue
```diff
@@ -20,35 +30,5 @@
...
-  name: 'CdrChip',
-  props: {
-    type: {
-      type: String,
-      validator: (value) => propValidator(
-        value,
-        ['info', 'warning', 'success', 'error'],
-      ),
-      default: 'info',
-    },
```
benjag, line 37: This prop was in there for some reason, it's not part of Chip's API

PR #25, opened 2022-01-26
refactor(tooltip): adds script setup syntax, adds emit options

benjag:
Also fixes an un-documentented issue with the popup mouse handlers not working

### DIFF
--- a/src/components/tooltip/CdrTooltip.vue
+++ b/src/components/tooltip/CdrTooltip.vue
@@ -1,3 +1,82 @@
+<script setup>
+import { useCssModule, useSlots, ref, onMounted, watch } from 'vue';
+import mapClasses from '../../utils/mapClasses';
+import CdrPopup from '../popup/CdrPopup';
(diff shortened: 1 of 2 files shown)


src/components/tooltip/CdrTooltip.vue
```diff
@@ -15,107 +94,10 @@
       :id="id"
       @closed="closeTooltip"
     >
-      <slot />
+      <slot ref="popupEl"/>
```
benjag, line 97: Left this in by mistake, removing...

src/components/tooltip/CdrTooltip.vue
```diff
@@ -1,3 +1,82 @@
...
+    type: String,
+    required: false,
+  },
+  open: {
+    type: Boolean,
+    default: false,
+  },
+});
+const slots = useSlots();
+const emit = defineEmits(['opened', 'closed']);
```
benjag, line 35: Here is an example of emit options

PR #35, opened 2022-02-23
fix(cdrcheckbox): add v-model bound to computed checkboxModel. Add cu…

KenjiCrosland:
…stom indeterminate directive

In order to properly update checkbox group, we had to bind v-model with a computed checkboxModel
with its own custom getter and setter. In addtion, since the 'indeterminate' property wasn't being
applied to the input regardless of what was being passed to :indeterminate, I created a custom
vIndeterminate which would add/remove the indeterminate prop from the checkbox input.

Also, there is an outstanding issue where arrays that are passed as values to checkbox groups are stringified and have the brackets removed. After some research I believe this to be a bug in vue itself. Since the use case seems very limited, I've created a story to look at it later: https://issues.rei.com/browse/CDR-2472

CDR-2458

### DIFF
--- a/src/components/checkbox/CdrCheckbox.vue
+++ b/src/components/checkbox/CdrCheckbox.vue
@@ -91,11 +91,11 @@ const props = defineProps({
  const emit = defineEmits(['update:modelValue'])
 
  const vIndeterminate = (el, binding)=>{
-    if(!binding.value) {
-      el.removeAttribute("indeterminate");
+    if(!!binding.value) {
+      el.setAttribute("indeterminate", binding.value)
       return;
     }
-    el.setAttribute("indeterminate", "indeterminate")
+    el.removeAttribute("indeterminate");
   }
  const checkboxModel = computed({
   get() {
--- a/src/components/checkbox/CdrCheckbox.vue
+++ b/src/components/checkbox/CdrCheckbox.vue
@@ -13,19 +13,24 @@
         :class="[style['cdr-checkbox__input'], inputClass]"
         type="checkbox"
         v-bind="$attrs"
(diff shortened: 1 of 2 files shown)


src/components/checkbox/CdrCheckbox.vue
```diff
@@ -13,19 +13,24 @@
...
         type="checkbox"
         v-bind="$attrs"
-        :checked="newValue"
-        @change="$emit('update:modelValue', $event.target.checked, $event)"
         :true-value="customValue ? null : trueValue"
         :false-value="customValue ? null : falseValue"
         :value="customValue"
-        :indeterminate="indeterminate"
-      >
+        v-indeterminate="indeterminate"
```
benjag, line 19: Not sure I like the use of `v-` here. To me this looks like like a vue-managed attribute and could be confusing ... also potentially breaking if they ever decided to use that themselves. Why not `isIndeterminate` or something like that?

src/components/checkbox/examples/checkboxes.vue
```diff
@@ -102,10 +102,10 @@
     </div>
 
     <cdr-checkbox
-      indeterminate
+      indeterminate="true"
     >indeterminate (not functional)</cdr-checkbox>
     <cdr-checkbox
-      indeterminate
+      indeterminate="true"
```
benjag, line 108: Does this break without the `="true"`?

src/components/checkbox/CdrCheckbox.vue
benjag: Disregard, I learned something new today!

benjag (approved):

PR #39, opened 2022-03-08
Pr/cdr 2153 refactor cdr input

KenjiCrosland:
- Refactor CdrInput to use <script setup> syntax. 
- Fixed v-model issue by referencing a new computed ‘inputModel’ with its own custom getter and setter. 
- Ensured all fallthrough attributes except class get applied to the input element. The class attribute is bound to the outermost element.
- Added a uid module to provide a unique id for each component should an id not be provided by the user. 
- Fixed some implementations issue on the CdrInput example page.
- Fixed a typo in CdrLabelStandalone.

### DIFF
--- a/src/components/formError/CdrFormError.vue
+++ b/src/components/formError/CdrFormError.vue
@@ -1,5 +1,5 @@
 <template>
-  <div :class="style[baseClass]" role="status" tabindex="0">
+  <div :class="style[baseClass]" role="status">
     <span :class="style[iconClass]">
       <icon-error-stroke size="small" inherit-color />
     </span>
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -224,8 +224,8 @@ const focusedClass = computed(() => isFocused.value && 'cdr-input--focus');
 
 const describedby = computed(() => {
   return [
-    slots['helper-text-top'] ? `${props.id}-helper-text-top` : '',
-    slots['helper-text-bottom'] ? `${props.id}-helper-text-bottom` : '',
+    slots['helper-text-top'] ? `${uniqueId}-helper-text-top` : '',
+    slots['helper-text-bottom'] ? `${uniqueId}-helper-text-bottom` : '',
     attrs['aria-describedby'],
   ].filter((x) => x).join(' ');
 })
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -1,11 +1,12 @@
 <template>
   <cdr-label-standalone
-    :for-id="id"
+    :for-id="uniqueId"
     :label="label"
     :hide-label="hideLabel"
     :required="required"
     :optional="optional"
     :disabled="disabled"
+    :class="attrs.class"
   >
     <template
       #helper
@@ -35,17 +36,16 @@
                            backgroundClass,
                            sizeClass,
         )"
-        :id="id"
+        :id="uniqueId"
         :disabled="disabled"
         :aria-required="required || null"
         :aria-invalid="!!error || null"
         :aria-errormessage="(!!error && `${id}-error`) || null"
         v-bind="inputAttrs"
-        :value="modelValue"
         :aria-describedby="describedby || null"
-        @input="$emit('update:modelValue', $event.target.value)"
         @focus="isFocused = true"
         @blur="isFocused = false"
+        v-model="inputModel"
       />
       <input
         v-else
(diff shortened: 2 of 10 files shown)


src/components/input/examples/Inputs.vue
```diff
@@ -377,29 +443,29 @@ export default {
   },
   watch: {
     $route(to) {
-      this.setBackground(to.query.background);
+      this.setBackground("primary");
     },
   },
   mounted() {
-    this.setBackground(this.$router.currentRoute.query.background);
+    this.setBackground("primary");
```
benjag, line 450: Just to note I think eventually this will need to be switched back once we re-enable this background toggle feature

KenjiCrosland: Oh good catch. I think I set it to a single value because the warnings were cluttering my console. I'm going to set it back

src/utils/uid.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Quickly generates a unique id string.
+ */
+
+export default function uid() {
+    let uid = Math.floor((1 + Math.random()) * 0x1000000).toString(16).substring(1);
```
benjag, line 6: Prefer `const`

mhewson:
pulling this down to do some screen reader validation

src/components/input/examples/Inputs.vue
```diff
@@ -172,7 +172,7 @@
         <span id="errorMessage">you have added too many characters, remove some</span>
```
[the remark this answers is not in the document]
KenjiCrosland: Good catch. I fixed the `undefined-error`.  It looks like `cdr-form-error` isn't structured so that the `id` of the span containing the error message has the same value as `aria-errormessage` on the `input` element. Rather it's referencing the div containing the span. Is that ok or do we need the id to be on the element that contains the text node of the error message itself?

Was this test run via `npm test`? I saw a11y tests pop up but I'm having chromedriver issues.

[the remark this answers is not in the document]
mhewson: No I manually tested this in Axe dev tools - you may need a license for that still, if so we can reach out to Eric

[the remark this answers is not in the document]
KenjiCrosland: Yes it does looks like I need a license for Axe Dev Tools. Last time I used it it was free but rather clunky IIRC

src/components/input/examples/Inputs.vue
```diff
@@ -125,19 +125,19 @@
...
         :background="backgroundColor"
       >
-        <template slot="pre-icon">
+        <template #pre-icon>
           <cdr-icon
             use="#twitter"
           />
         </template>
-        <template slot="post-icon">
+        <template #post-icon>
```
mhewson, line 133: This use-case seems to have a tab focus problem, rendered example label "Form with Two Buttons *"
tabbing into this group places focus on the entire input, tabbing a second time does not move focus to the actual input, nor does it place a cursor into the input field, tabbing a 3rd time moves focus to the close button

src/components/input/examples/Inputs.vue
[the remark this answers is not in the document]
mhewson: OH and this will need to be a unique ID, not "errormessage" but maybe {{input-lable}}-errormessage or something?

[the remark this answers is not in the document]
KenjiCrosland: Yes I fixed that. I got a trial account of axe and it looks like the issue isn't popping up. Will double check after office hours

src/components/input/examples/Inputs.vue
KenjiCrosland: Form error has been updated to remove `tab-index=0` from the form error components

src/components/formError/CdrFormError.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div :class="style[baseClass]" role="status" tabindex="0">
```
KenjiCrosland, line 2: This fixes the accessibility issue where the error was tabbable.

src/components/input/examples/Inputs.vue
```diff
@@ -169,10 +169,10 @@
       </template>
 
       <template #error>
-        <span id="errorMessage">you have added too many characters, remove some</span>
+        <span>you have added too many characters, remove some</span>
```
KenjiCrosland, line 172: This id is superfluous and is the cause of some axe accessibility issues.

src/components/labelStandalone/CdrLabelStandalone.vue
```diff
@@ -8,7 +8,6 @@
         {{ label }}{{ required || optional ? '' : '' }}
         <span
           v-if="required"
-          aria-label="required"
```
KenjiCrosland, line 11: `aria-label="required"` is not needed here. Since the input element already has `aria-required=true` all this does is make screen readers say "Required" twice.

mhewson (approved):

src/components/input/CdrInput.vue
```diff
@@ -117,29 +116,30 @@
...
   </cdr-label-standalone>
 </template>
+
 <script>
-import { defineComponent, useCssModule, computed, ref } from 'vue';
+export default {
+  inheritAttrs: false,
+  customOptions: {}
+}
+</script>
```
benjag, line 125: Interesting, I was wondering when we might run in to needing to do this with the setup syntax

benjag (approved):

PR #45, opened 2022-03-22
refactor(cdr-select): refactor cdr-select to use Vue 3 <script setup>…

KenjiCrosland:
… syntax

- Add v-model bound to computed selectModel value.
- Add unique id. 
- Remove toArray import from lodash.
- Remove function which computed selected options for a multiple select (v-model takes care of this).
- Fix styling issues with incorrect classes being applied.

### DIFF
--- a/src/components/select/examples/Selects.vue
+++ b/src/components/select/examples/Selects.vue
@@ -139,11 +139,11 @@
 
       prompt="Choose One"
     >
-      <template slot="helper-text">
+      <template #helper-text>
         This is helper text.
       </template>
 
-      <template slot="info">
+      <template #info>
         <cdr-link
           href="#/selects"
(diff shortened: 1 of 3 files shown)


src/components/select/examples/Selects.vue
```diff
@@ -241,14 +242,35 @@
...
+      @change="validatePreIconModel2"
     >
-      <template slot="error">
+      <template #error>
         <span id="alertTest">Alert error message goes here</span>
       </template>
     </cdr-select>
     <hr class="icon-hr">
 
+    <!-- Nested Options -->
```
KenjiCrosland, line 254: This was a docsite example that didn't exist in the kitchen sink

benjag (approved):

PR #46, opened 2022-03-24
refactor(cdr-table): refactor cdr-table to use <script setup> syntax

KenjiCrosland:
CDR-2511


src/components/table/CdrTable.vue
```diff
@@ -1,3 +1,53 @@
+<script setup>
```
KenjiCrosland, line 1: No curveballs here. Just a simple refactor.

mhewson (approved):

PR #48, opened 2022-04-06
refactor(cdr-pagination): cdr-pagination <script setup> refactor

KenjiCrosland:
Remove example with slotted links. Fix issues with simulated clicks from option select. Add
additional test for select. Add uid module of ids. Fix snapshots. Fix Prev and Next link text.

CDR-2516

### DIFF
--- a/src/components/pagination/CdrPagination.vue
+++ b/src/components/pagination/CdrPagination.vue
@@ -1,6 +1,6 @@
 
 <script setup>
-import { useCssModule, defineEmits, computed, ref, watch, nextTick, onMounted } from 'vue';
+import { useCssModule, computed, ref, watch, nextTick, onMounted } from 'vue';
 import mapClasses from '../../utils/mapClasses';
 import propValidator from '../../utils/propValidator';
 import IconCaretLeft from '../icon/comps/caret-left';
--- a/src/components/pagination/CdrPagination.vue
+++ b/src/components/pagination/CdrPagination.vue
@@ -1,118 +1,17 @@
-<template>
-  <nav :aria-label="ariaLabel">
-    <ol :class="style['cdr-pagination']">
-      <li v-if="innerValue > pages[0].page">
-        <component
-          :is="linkTag"
-          aria-label="Go to previous page"
-          :href="(linkTag === 'a' && prevPageData && prevPageData.url) || undefined"
-          :class="mapClasses(style, 'cdr-pagination__link', 'cdr-pagination__prev')"
-          @click="(e) => navigate(prevPageData.page, e)"
-        >
-          <icon-caret-left :class="style['cdr-pagination_caret--prev']" />
-        </component>
-      </li>
-      <li v-else aria-hidden="true">
-        <span
-          aria-disabled="true"
-          :class="mapClasses(style, 'cdr-pagination__link', 'cdr-pagination__prev', 'cdr-pagination__link--disabled')"
-        >
-          <icon-caret-left
-            :class="style['cdr-pagination_caret--prev']"
-            inherit-color
-          />
-        </span>
-      </li>
 
-
-      <li
-        v-for="n in paginationData"
-        :key="`pagination-${id}-li-${n.page}`"
-        :class="style['cdr-pagination__li--links']"
-      >
-
-        <component
-          v-if="n.page"
-          :is="linkTag"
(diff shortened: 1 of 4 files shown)


src/components/pagination/CdrPagination.vue
```diff
@@ -49,15 +220,15 @@
 
       <li :class="style['cdr-pagination__li--select']">
         <cdr-select
-          :id="`pagination-select-${id}`"
+          :id="`pagination-select-${uniqueId}`"
           v-model="innerValue"
           label="Navigate to page"
           hide-label
-          @change="select"
+          @change.prevent="(e) => select(e)"
```
KenjiCrosland, line 227: We can't use the `navigate` function directly here because we need to simulate a click on the corresponding page link element to in order to get some default click behavior that is not included in `navigate`

src/components/pagination/CdrPagination.vue
```diff
@@ -1,3 +1,170 @@
...
+        const target = e.currentTarget || e.target;
+        target.blur();
+      } catch (err) {
+        // eslint-disable-next-line no-console
+        console.error(err);
+      }
+    });
+  }
+
+  const select = (e) =>{
```
KenjiCrosland, line 104: This function could be greatly simplified because we no longer use slots for pagination.

src/components/pagination/CdrPagination.vue
```diff
@@ -1,3 +1,170 @@
...
+  });
+
+  const setCurrentIdx = (page) => {
+    currentIdx.value = props.pages.map((x) => x.page).indexOf(page);
+  }
+
+  const navigate = (pageNum, e) => {
+    // TODO: provide option to just emit without navigating to support vue-router?
+    // Dont do anything if clicking the current active page
+    const paginationLinkIsVisible = e.target.offsetWidth > 0 && e.target.offsetHeight > 0;
```
KenjiCrosland, line 82: When the pagination link is visible, we should expect the default behavior when someone clicks the pagination link for the current page.

The pagination link will not be visible, however, when we're using the select element instead of a list of links (this is in mobile view widths). In this case a click is simulated on the hidden link and we need the navigate event to emit.

src/components/pagination/__tests__/CdrPagination.spec.js
```diff
@@ -281,20 +281,40 @@ describe('CdrPagination', () => {
     expect(wrapper.emitted().navigate[5][0]).toBe(2);
     expect(wrapper.emitted().navigate[5][1]).toBe('?page=2');
     expect(wrapper.emitted().navigate[5][2] instanceof Event).toBeTruthy();
+    
     // Do nothing when clicking current page link
+    const originalOffsetHeight = Object.getOwnPropertyDescriptor(HTMLElement.prototype,'offsetHeight');
```
KenjiCrosland, line 286: This felt a bit clunky, but this is what stack overflow recommended as a way to provide offsetHeight and width for the pagination links. It appears in that in this environment they are either zero or undefined.

src/components/pagination/__tests__/CdrPagination.spec.js
```diff
@@ -281,20 +281,40 @@ describe('CdrPagination', () => {
...
-    // expect(wrapper.emitted().navigate[6][1]).toBe('?page=3');
-    // expect(wrapper.emitted().navigate[6][2] instanceof Event).toBeTruthy();
+    //The event should be emitted when the link is not visible (simulated click after making a selection)
+    link2.trigger('click'); // 2 -> 2
+    await wrapper.vm.$nextTick();
+    expect(wrapper.emitted().navigate[6][0]).toBe(2);
+    expect(wrapper.emitted().navigate[6][1]).toBe('?page=2');
+    expect(wrapper.emitted().navigate[6][2] instanceof Event).toBeTruthy();
+
+    //Options select
```
KenjiCrosland, line 311: New test for option select added.

src/components/pagination/CdrPagination.vue
```diff
@@ -1,3 +1,170 @@
+
+<script setup>
+import { useCssModule, defineEmits, computed, ref, watch, nextTick, onMounted } from 'vue';
```
benjag, line 3: Small nitpick, you don't need to import `defineEmits`, it's a compiler macro like `defineProps`

benjag (approved):

PR #57, opened 2022-04-27
fix(cdr-popover): fix popover focus bug and custom trigger open state…

KenjiCrosland:
… bug

Added longer timeout to fix focus issue. Added custom handler to the "close" event to ensure that
the open state wasn't set to true after the popover with the custom trigger closed

### DIFF
--- a/src/components/popover/CdrPopover.vue
+++ b/src/components/popover/CdrPopover.vue
@@ -108,10 +108,10 @@
     lastActive = activeElement;
     isOpen.value = true;
     emit('opened', e);
-    nextTick(() => {
+    setTimeout(()=>{
       const tabbables = tabbable(popupEl.value.$el);
       if (tabbables[0]) tabbables[0].focus();
-    });
+    }, 50)
   }
 
   const closePopover = (e) => {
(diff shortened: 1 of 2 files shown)


src/components/popover/examples/Popover.vue
```diff
@@ -159,6 +158,10 @@ export default {
     popupHandler(e) {
       console.log(e);
     },
+    closedCustomHandler(e) {
```
KenjiCrosland, line 161: Every popover with a custom trigger will need to have to listen for the "closed" event after the popover is closed, because popovers with custom triggers have no control over the state of the button that controls them. This is actually broken in the docsite too and will need an extra bullet point to let users know that they need to add an event handler in order for it to work properly

src/components/popover/CdrPopover.vue
```diff
@@ -108,10 +108,10 @@
     lastActive = activeElement;
     isOpen.value = true;
     emit('opened', e);
-    nextTick(() => {
+    setTimeout(()=>{
       const tabbables = tabbable(popupEl.value.$el);
       if (tabbables[0]) tabbables[0].focus();
-    });
+    }, 50)
```
KenjiCrosland, line 114: Turns out we just needed a bit more time

benjag: Don't we all

benjag (approved):

PR #62, opened 2022-05-04
fix(cdr-select): remove role="status" default for errors. Add error-i…

KenjiCrosland:
…d to aria-describedby attr

CDR-2355

### DIFF
--- a/src/components/select/CdrSelect.vue
+++ b/src/components/select/CdrSelect.vue
@@ -60,7 +60,6 @@ const props = defineProps({
   errorRole: {
     type: String,
     required: false,
-    default: '',
   },
   modelValue: {
     type: [String, Number, Boolean, Object, Array, Symbol, Function],
--- a/src/components/select/CdrSelect.vue
+++ b/src/components/select/CdrSelect.vue
@@ -60,7 +60,7 @@ const props = defineProps({
   errorRole: {
     type: String,
     required: false,
-    default: 'status',
+    default: '',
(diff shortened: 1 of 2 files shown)


src/components/select/CdrSelect.vue
```diff
@@ -60,7 +60,7 @@ const props = defineProps({
   errorRole: {
     type: String,
     required: false,
-    default: 'status',
+    default: '',
```
KenjiCrosland, line 63: Removed this as a default because the error got announced twice

src/components/select/examples/Selects.vue
```diff
@@ -154,7 +154,34 @@
     </cdr-select>
     <cdr-text>Selected Value: {{ helperTextModel }}</cdr-text>
     <hr class="icon-hr">
+    
+    <cdr-select
```
KenjiCrosland, line 158: Added example with both helper text and an error.

src/components/select/CdrSelect.vue
benjag: Should the default be `undefined` then instead of an empty string?

mhewson: taking a look

KenjiCrosland: > Should the default be `undefined` then instead of an empty string?

Updated to remove the default property.

mhewson (approved):

PR #64, opened 2022-05-09
feat(cedar e2e tests): move all nightwatch globals to nightwatch.conf…

KenjiCrosland:
….js. Delete globals.js file

Move all configs from the globals.js file to nightwatch.conf.js. Remove unused nightwatch-vite
plugin.

CDR-2520


benjag (approved):
I'll follow up with a PR to fix that speed issue

PR #68, opened 2022-05-16
test(cdr-accordion and cdr-accordion-group): refactor CdrAccordion an…

KenjiCrosland:
…d CdrAccordionGroup tests

CDR-2541

### DIFF
--- a/src/components/accordion/__tests__/CdrAccordion.spec.js
+++ b/src/components/accordion/__tests__/CdrAccordion.spec.js
@@ -1,212 +1,158 @@
 import { shallowMount, mount } from '../../../../test/vue-jest-style-workaround.js';
 import CdrAccordion from '../CdrAccordion.vue';
 
-const propsData = {
-  id: 'test',
-  label: 'A label',
-  level: '2',
-  compact: false,
-  borderAligned: false,
-};
+const baseComponentPattern = {
+  propsData: {
+    id: 'test',
+    level: '2',
+  },
+  slots: {
+    default: 'This is some slot text.',
+    label: 'label',
+  },
+}
 
 describe('CdrAccordion', () => {
-  it('renders correctly', () => {
-    const wrapper = mount(CdrAccordion, {
-      propsData: {
-        id: 'test',
-        level: '2',
-      },
-      slots: {
-        default: 'This is some slot text.',
-        label: 'label'
-      },
-    });
-    expect(wrapper.element).toMatchSnapshot()
-  });
+  describe('when mounted', () => {
+    let wrapper;
 
-  it('renders correctly unwrapped', async () => {
-    const wrapper = mount(CdrAccordion, {
-      propsData: {
-        id: 'test',
-        level: '2',
-      },
-      slots: {
-        default: 'This is some slot text.',
-        label: 'label'
-      },
-      global: {
-        provide: {
-          unwrap: {
-            isUnwrapped: {
-              value: true
-            }
-          }
-        }
-      }
+    beforeEach(() => {
+      wrapper = mount(CdrAccordion, { ...baseComponentPattern });
     });
-    await wrapper.vm.$nextTick();
-    expect(wrapper.element).toMatchSnapshot()
-  });
 
-  // TODO add more checks here
-  it('meets a11y requirements', async () => {
-    const wrapper = shallowMount(CdrAccordion, {
-      propsData: {
-        id: 'test',
-        level: '2',
-      },
-      slots: {
-        default: 'This is some slot text.',
-        label: 'label',
-      },
+    it('renders correctly', () => {
+      expect(wrapper.element).toMatchSnapshot()
     });
-    const button = wrapper.find('#test');
-    const contentArea = wrapper.find('.cdr-accordion__content');
-    expect(button.classes()).toContain('js-cdr-accordion-button');
-    // closed state
-    expect(button.attributes('aria-expanded')).toBe('false');
-    expect(button.attributes('aria-controls')).toBe(`${wrapper.vm.id}-collapsible`);
-
-    // opened state
-    button.trigger('click');
-    wrapper.setProps({ opened: true }); // fake the opening logic
-    await wrapper.vm.$nextTick();
-    expect(button.attributes('aria-expanded')).toBe('true');
-    expect(contentArea.attributes('aria-hidden')).toBe('false');
   })
 
-  describe('accordion data setup', () => {
-    it('sets maxHeight when starting closed', () => {
-      const wrapper = shallowMount(CdrAccordion, {
-        propsData: {
-          id: 'test',
-          level: '2',
-          opened: false,
-        },
-        slots: {
-          default: 'This is some slot text.',
-          label: 'label',
-        },
+  describe('when unwrapped', () => {
+    let wrapper;
+
+    beforeEach(() => {
+      const unwrappedComponent = {
+        ...baseComponentPattern,
+        global: {
+          provide: {
+            unwrap: {
+              isUnwrapped: {
+                value: true
+              }
+            }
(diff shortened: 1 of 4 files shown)


src/components/accordion/__tests__/CdrAccordionGroup.spec.js
```diff
@@ -3,79 +3,95 @@ import CdrAccordionGroup from '../CdrAccordionGroup.vue';
...
+    describe('a11y requirements', ()=>{
+      it('accordion groups are organized into unordered lists', ()=>{
+        expect(wrapper.element.tagName).toBe('UL');
+      })
+    });
   });
+
+  //TODO: Since the keyboard interaction tests don't work here, we may want write new e2e tests to ensure keyboard interactions work
+
+  // it('has correct a11y', async () => {
```
KenjiCrosland, line 47: Since this doesn't really test anything, I'm commenting it out. However, when we add more e2e tests these old tests can be useful as a reference for writing the new ones.

src/components/accordion/__tests__/CdrAccordion.spec.js
```diff
@@ -1,212 +1,158 @@
...
+          expect(button.attributes('aria-controls')).toBe(`${wrapper.vm.id}-collapsible`);
+        });
+      })
+    })
+
+    describe('when the accordion is open', () => {
+      beforeEach(async() => {
+        button.trigger('click');
+        wrapper.setProps({ opened: true }); // fake the opening logic
+        //TODO: e2e test where clicks work as expected. These could be visual regression tests where we check the visual appearance of the open state
```
KenjiCrosland, line 113: This is another case for an e2e test that we can add. We aren't testing the implementation of the click, just what happens when the `opened` prop changes. We're just assuming that a click will do that in this test.

src/components/accordion/__tests__/__snapshots__/CdrAccordionGroup.spec.js.snap
```diff
@@ -189,6 +189,41 @@ exports[`CdrAccordionGroup > renders correctly unwrapped 1`] = `
       >
         
         
+      </div>
```
KenjiCrosland, line 192: I added a new accordion tab to the snapshot to consolidate the code, so the snapshot has changed

src/components/accordion/__tests__/__snapshots__/CdrAccordion.spec.js.snap
```diff
@@ -1,6 +1,6 @@
 // Vitest Snapshot v1
 
-exports[`CdrAccordion > renders correctly 1`] = `
+exports[`CdrAccordion > when mounted > renders correctly 1`] = `
```
KenjiCrosland, line 3: When the test structure changes but the markup remains the same, it becomes "obsolete". Interestingly the snapshots still work as expected

src/components/accordion/__tests__/CdrAccordion.spec.js
```diff
@@ -1,212 +1,158 @@
 import { shallowMount, mount } from '../../../../test/vue-jest-style-workaround.js';
```
KenjiCrosland, line 1: Some general formatting changes made here:

All expects statements should correspond to an it block
All test setup should not be in an it block, but in the before hook of a describe block.
Use nested describes when it makes sense

src/components/accordion/__tests__/CdrAccordion.spec.js
```diff
@@ -1,212 +1,158 @@
...
+          });
+          it('should NOT be focused', ()=>{
+            expect(wrapper.vm.focused).toBeFalsy();
+          })
+          it('should NOT have the proper focus class', ()=>{
+            expect(wrapper.classes()).not.toContain('cdr-accordion--focused');
+          })
+        });
+      })
+      describe('a11y requirements', () => {
```
KenjiCrosland, line 90: There are two separate describes for "a11y requirements". I think it makes sense to have top level describes for the state of a component, and nested "a11y requirements" for each one

src/components/accordion/__tests__/CdrAccordion.spec.js
```diff
@@ -1,212 +1,158 @@
...
-        propsData: {
-          id: 'test',
-          level: '2',
-          opened: false,
-        },
-        slots: {
-          default: 'This is some slot text.',
-          label: 'label',
-        },
+  describe('when shallow mounted', () => {
```
benjag, line 51: 'when shallow mounted' feels odd to me as a describe since that's a specific pattern for a test and not how this would be used in real life

src/components/accordion/__tests__/CdrAccordion.spec.js
```diff
@@ -1,212 +1,158 @@
...
+    })
+
+    describe('when the accordion is open', () => {
+      beforeEach(async() => {
+        button.trigger('click');
+        wrapper.setProps({ opened: true }); // fake the opening logic
+        //TODO: e2e test where clicks work as expected. These could be visual regression tests where we check the visual appearance of the open state
+      })
+      it('updates maxHeight on prop update', () => {
+        expect(wrapper.vm.maxHeight).toBe(0); //TODO: Investigate. Shouldn't this value be above zero? --Kenji
```
benjag, line 116: Yeah, `max-height` is set to 0 when closed and `none` when opened

KenjiCrosland: Figured this out. The property only updates after a 400ms timeout, so I put a 500ms timeout in the test.

src/components/accordion/__tests__/CdrAccordion.spec.js
KenjiCrosland: That makes sense. After looking into the difference between `mount` and `shallowMount` I think it's fair to describe these a component unit tests.

src/components/accordion/__tests__/CdrAccordionGroup.spec.js
```diff
@@ -3,79 +3,95 @@ import CdrAccordionGroup from '../CdrAccordionGroup.vue';
...
+    });  
   });
 
-  it('has correct a11y', async () => {
-    // const elem = document.createElement('div')
-    // if (document.body) {
-    //   document.body.appendChild(elem)
-    // }
-    //   attachTo: elem,
+  describe('when unwrapped', () => {
```
KenjiCrosland, line 22: @benjag your comment on the other describe makes me rethink this describe as well. Maybe we should separate our snapshot tests from our unit tests? In that case, I'd move the one test for a11y requirements into its own describe and mount the same component. Feels a bit redundant but may be good for separation of concerns

src/components/accordion/__tests__/CdrAccordion.spec.js
```diff
@@ -1,212 +1,161 @@
...
-    const wrapper = mount(CdrAccordion, {
-      propsData: {
-        id: 'test',
-        level: '2',
-      },
-      slots: {
-        default: 'This is some slot text.',
-        label: 'label'
-      },
+  describe('component snapshot when wrapped', () => {
```
KenjiCrosland, line 16: Update to be more explicit

src/components/accordion/__tests__/CdrAccordion.spec.js
```diff
@@ -1,212 +1,161 @@
...
-      slots: {
-        default: 'This is some slot text.',
-        label: 'label'
-      },
-      global: {
-        provide: {
-          unwrap: {
-            isUnwrapped: {
-              value: true
+  describe('component snapshot when unwrapped', () => {
```
KenjiCrosland, line 28: Update to be more explicit

benjag (approved):

mhewson (approved):

PR #75, opened 2022-05-25
Pr/cdr 2541 refactor button tests

KenjiCrosland:


src/components/button/__tests__/CdrButton.spec.js
```diff
@@ -3,113 +3,135 @@ import CdrButton from '../CdrButton.vue';
...
+    describe('with an icon-left slot', () => {
+      beforeEach(() => {
+        wrapper = shallowMount(CdrButton, {
+          slots: {
+            default: ['default slot for text'],
+            'icon-right': '<div />',
+          },
+        });
+      });
+      it('adds the "cdr-button--has-icon-right" class', ()=>{
```
KenjiCrosland, line 132: Should we do a snapshot test here, instead of testing for just one class? That way we can account for any other outliers.

There are a lot of tests that are testing for one specific attribute or class. Maybe a snapshot test should be added in addition to these. While this is redundant, coverage-wise, sometimes specific expect statements can be useful in calling out what's wrong so that we don't have to dig into the snapshot comparison every time when there's a failing unit test

benjag (approved):

PR #88, opened 2022-06-13
test(cdr-modal tests): refactor cedar modal tests

KenjiCrosland:
CDR-2541

### DIFF
--- a/src/components/modal/__tests__/CdrModal.spec.js
+++ b/src/components/modal/__tests__/CdrModal.spec.js
@@ -47,7 +47,7 @@ describe('CdrModal.vue', () => {
   });
 
 
-  describe('', ()=>{
+  describe('default closed', ()=>{
     let wrapper;
     let elem;
     beforeEach(()=>{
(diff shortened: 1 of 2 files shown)


src/components/modal/__tests__/__snapshots__/CdrModal.spec.js.snap
```diff
@@ -73,7 +73,7 @@ exports[`CdrModal.vue > default open 1`] = `
             <div
               class="cdr-modal__text-content"
               role="document"
-              style="max-height: -80px; padding-right: 0px;"
```
KenjiCrosland, line 76: @benjag this looks right? We should see a positive value for the opened state correct?

Also, maybe we should have another snapshot for a closed state?

benjag: Ok, I expect you're right on the nextTick() being too fast then

KenjiCrosland: I think this is good to go. I added another snapshot for the closed state.

src/components/modal/__tests__/CdrModal.spec.js
```diff
@@ -26,6 +26,10 @@ describe('CdrModal.vue', () => {
       expect(wrapper.element).toMatchSnapshot();
```
KenjiCrosland, line 26: @benjag I noticed these tests were commented out and your name was on the blame. I re-enabled them in this PR, fyi

benjag (approved):

PR #89, opened 2022-06-15
Fix lint config

benjag:
Our lint config hasn't been working correctly since the vite conversion. This should fix it. The diff for the eslint file is pretty crazy because it switched line ending style from CLRF to LF


benjag:
This one isn't quite ready for review yet, couple more commits incoming

benjag:
ESlint really isn't there yet with fully supporting the <script setup> sugar. Future tech topic discussion methinks

PR #102, opened 2022-08-18
Image container fix

benjag:
CdrImg has always bound any pass-through attributes to the  element itself. This allows a consumer to pass something like `loading="lazy"`. 

Vue 2 did not include `class` or `style` attributes as part of the `$attrs` object, so they would be treated as default attribute pass-throughs and be applied to the parent element of CdrImg. The attribute object has been changed in Vue 3 to include class and style and thus they now pass down to the img element as weill.
https://v3-migration.vuejs.org/breaking-changes/attrs-includes-class-style.html

Overall this leads to a more consistent experience, however many of our consumers relied on the interaction of passing a class to the parent element, particularly when being used with a ratio (which adds a wrapping ratio container div).

This PR introduces a prop to allow passing a class to that element, while still allowing for the updated interaction of class/style attributes being passed to the `img` element.

### DIFF
--- a/src/components/image/CdrImg.vue
+++ b/src/components/image/CdrImg.vue
@@ -2,7 +2,7 @@
   <div
     v-if="ratio"
     :style="ratioObject"
-    :class="style[ratioClass]"
+    :class="[style[ratioClass], containerClass]"
   >
     <img
       :style="cropObject"
@@ -85,6 +85,10 @@ const props = defineProps({
   cover: {
     type: Boolean,
   },
+    /**
+   * Adds a custom class to the cdr-img__ratio container div
+   */
+  containerClass: String,
   /**
    * Sets a border radius to an element {square, top, right, bottom, left}
    */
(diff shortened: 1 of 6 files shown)


src/components/image/CdrImg.vue
```diff
@@ -85,6 +85,10 @@ const props = defineProps({
   cover: {
     type: Boolean,
   },
+    /**
+   * Adds a custom class to the cdr-img__ratio container div
+   */
+  containerClass: String,
```
mhewson, line 91: would it make it more clear to call this ratioContainerClass or this added for all modifiers?

HeavyMedl (approved):

mhewson (approved):

src/components/image/CdrImg.vue
benjag: My thinking was that consumers are using this class in ways unrelated to the ratio styles, so it felt a little more confusing to me to scope it as such

PR #107, opened 2022-08-31
When server-rendered, `cdr-accordion` shows the height of the content within the content container on load

HeavyMedl:
Discovered a side-effect of server-rendering the accordion while converting the `curbside-pickup` application in `landing-site`.

In the context of SSR, the `watch` callback doesn't execute unless we force it with `{immediate: true}`. Rather than do this, we can set an initial value on `maxHeight.value` which solves the issue.

Screen Shot 2022-08-31 at 2 22 42 PM


src/components/accordion/__tests__/CdrAccordion.spec.js
```diff
@@ -59,7 +59,7 @@ describe('CdrAccordion', () => {
     })
     describe('when the accordion is closed', () => {
       it('sets maxHeight to be 0', () => {
-        expect(wrapper.vm.maxHeight).toBe(0);
+        expect(wrapper.vm.maxHeight).toBe('0');
```
HeavyMedl: This had to be updated to the String "0". Something about it being the number 0 was making this not compute in the context of the SSR bundle. Internally its probably doing some kind of `if` logic on the value and 0 is computing as `false`.

Additionally, we might want to normalize this to always append the unit measurement to the string e.g. `0px` to be consistent

benjag (approved):

benjag:
Thanks Kurt!

PR #114, opened 2022-09-26
fix(modal): fix rendering of close button when title is hidden

benjag:
This reverts some of the "flattening" I did in modal, which inadvertently caused a small visual issue when the title was hidden.

Will go out as patch version 13.0.2

### DIFF
--- a/src/components/modal/styles/CdrModal.module.scss
+++ b/src/components/modal/styles/CdrModal.module.scss
@@ -79,13 +79,13 @@ $modal-animation-duration: 150ms;
   &__header {
     display: flex;
     padding-bottom: $cdr-space-one-x;
-    flex: auto;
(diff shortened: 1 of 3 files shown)


src/components/modal/styles/CdrModal.module.scss
```diff
@@ -79,13 +79,13 @@ $modal-animation-duration: 150ms;
...
-    flex: auto;
     justify-content: space-between;
+  }
+  &__title {
+    flex: auto;
     @include cdr-text-heading-serif-600;
   }
   &__close-button {
-    align-self: flex-start;
-    border: 1px solid $cdr-color-icon-default;
```
benjag, line 88: These rules weren't doing anything

KenjiCrosland (approved):

PR #118, opened 2022-11-11
feat(skeleton): add skeleton component

benjag:


src/components/skeleton/styles/CdrSkeletonBone.module.scss
```diff
@@ -0,0 +1,57 @@
+@use "sass:math";
```
benjag: @TODO - delete

PR #126, opened 2023-02-15
feat(input): adds support for type of date

benjag:
Unlocks the ability to pass `type="date"` and associated attributes to the underlying input. Unfortunately there's no cross-browser standard for styling the parts of the native date-picker so the experiences is not going to be consistent across browsers. 

That being said, I do think this unlocks enough utility for internal applications to warrant the change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@rei/cedar",
-  "version": "13.2.2",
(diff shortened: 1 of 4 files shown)


benjag:
Would go out as 13.3.0

mhewson (approved):

PR #127, opened 2023-03-02
Feat/add docgen

benjag:
Adds docgen capability and component documentation


old-docgen.mjs
```diff
@@ -0,0 +1,124 @@
+import { parse } from 'vue-docgen-api'
```
KenjiCrosland: @benjag I wonder if we still need this file. probably not

src/components/accordion/CdrAccordion.vue
```diff
@@ -189,6 +195,10 @@ export default defineComponent({
       :is="headingTag"
       :class="style[headingClass]"
     >
+      <!--
+        Triggered on accordion open/close
+        @event accordion-toggle
+      -->
```
benjag, line 201: TODO - delete (covered in `emits` object)

KenjiCrosland (approved):

mhewson (approved):

PR #128, opened 2023-03-14
Modal fix

benjag:
Resolves issue where a long title would cause the flex container to grow (and the button would stretch vertically)

### DIFF
--- a/src/components/modal/styles/CdrModal.module.scss
+++ b/src/components/modal/styles/CdrModal.module.scss
@@ -87,6 +87,7 @@ $modal-animation-duration: 150ms;
   }
   &__close-button {
     flex: none;
+    align-self: center;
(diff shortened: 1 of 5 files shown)


src/components/modal/styles/CdrModal.module.scss
```diff
@@ -87,6 +87,7 @@ $modal-animation-duration: 150ms;
   }
   &__close-button {
     flex: none;
+    align-self: center;
```
benjag, line 90: This is the relevant fix

mhewson (approved):

src/components/modal/styles/CdrModal.module.scss
mhewson: bah

PR #129, opened 2023-03-15
switch component from button to using an input type checkbox

stackoverfloweth:
## Description
rei-cedar team, as a fellow frontend engineer who recently built a vue3 component library I'm always super excited to find another company with engineers who chose to write their own instead of using vutify or primevue, so props for that.

My curiosity lead to finding your switch component, which I was surprised to see just used a button and not a checkbox under the hood. I was under the opinion that a checkbox was the ideal element to store the state for a switch, though in my testing the accessibility and voiceover are virtually identical. I did see that [both are acceptable patterns](https://www.w3.org/WAI/ARIA/apg/patterns/switch/)

> If the switch element is an HTML input[type="checkbox"], it uses the HTML checked attribute instead of the aria-checked property.

Unfortunately using a label/checkbox is not focusable by default, and pressing "space" doesn't toggle the value like it does a button. Obviously those are easy problems to solve and are solved in this PR, but makes the argument for switching to a checkbox a bit harder to justify.

I also wasn't able to figure out how to update snapshots for your snapshot testing with jest, so apologies for that.

Anyways, kudos on the component library 👍

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] iOS
- [ ] Android

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards

### Documentation:

### DIFF
--- a/src/components/switch/CdrSwitch.vue
+++ b/src/components/switch/CdrSwitch.vue
@@ -38,9 +38,6 @@ export default defineComponent({
   setup(props, ctx) {
     const style = useCssModule();
     const uniqueId = props.id ? props.id : uid();
-    const onClick = (modelValue) => {
-      ctx.emit('update:modelValue', !modelValue);
-    };
     const baseClass = 'cdr-switch';
     const sizeClass = computed(() => (props.size
       ? `cdr-switch--${props.size}`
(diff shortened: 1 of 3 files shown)


benjag (commented):
Hey @stackoverfloweth - Thanks so much for taking a look at our library. In the context of our [design system](https://rei.github.io/rei-cedar-docs/), we still feel that a `<button>` implementation is the most semantically correct option. 

Our intention with this component is to truly serve as an "on/off" toggle, we prefer our users to use CdrCheckbox when dealing with checklist-type UI. You can read more about our thinking in the [guidelines section](https://rei.github.io/rei-cedar-docs/components/switch/#guidelines) for switch.

PR #130, opened 2023-04-26
Update accessibility tests

benjag:
- Brings in the latest version of nightwatch and switches the axe-rules to the provided `axe-rules-verbose` plugin.
- Made a handful of changes to the integration examples due to new rules


src/components/breadcrumb/examples/Breadcrumb.vue
```diff
@@ -9,25 +9,33 @@
     <cdr-breadcrumb
       :items="averageBreadcrumbItems"
       data-backstop="breadcrumbs-ellipsis"
+      aria-label="two-item-breadcrumb"
+      id="two-item-breadcrumb"
```
benjag, line 13: Two issues here:

1. If an `id` is not passed, one is auto-generated at runtime. Passing an id ensures it's unique.
2. The parent element of a breadcrumb is a `nav`, when multiple breadcrumb components are on the page (not standard), the `nav` landmark is not uniquely labelled. Passing an aria-label solves that. 

https://github.com/rei/rei-cedar/blob/main/src/components/breadcrumb/CdrBreadcrumb.vue#L75-L78

src/components/pagination/examples/Pagination.vue
```diff
@@ -10,6 +10,7 @@
       :pages="pages"
       :total-pages="10"
       data-backstop="pagination-default"
+      for-label="normal"
```
benjag, line 13: Similar to breadcrumb, this prop also assigns an aria-label to the parent `nav` element. Passing one ensures it is unique.
https://github.com/rei/rei-cedar/blob/main/src/components/pagination/CdrPagination.vue#L186

src/components/checkbox/examples/checkboxes.vue
```diff
@@ -135,7 +135,6 @@
         Select All
       </cdr-checkbox>
       <cdr-list
-        role="group"
```
benjag, line 138: I think this should technically still be valid but Axe didn't like it so removing the group role.

src/components/quote/examples/demo/Pullquotes.vue
```diff
@@ -11,6 +11,7 @@
       <cdr-quote
         tag="aside"
         modifier="pull"
+        aria-label="quote-1"
```
benjag, line 14: Given the `aside` landmark a unique label

src/components/radio/examples/Radios.vue
```diff
@@ -6,32 +6,25 @@
         <legend id="legend-size">
           Radio Button Size Group
         </legend>        
-        <cdr-list aria-labelledby="legend-size" role="radiogroup">
-          <li>
+        <div aria-labelledby="legend-size" role="radiogroup">
```
benjag, line 9: A list did not seem like the correct markup here, so changing to a div

src/components/table/examples/Table.vue
```diff
@@ -442,7 +442,7 @@
             </th>
           </tr>
           <tr>
-            <th />
+            <th>-</th>
```
benjag, line 445: Axe does not like empty table header cells

src/components/tabs/examples/Tabs.vue
```diff
@@ -20,7 +20,7 @@
         <cdr-tab-panel
           v-for="tab in tabs.slice(0, 5)"
           :key="tab"
-          :name="tab"
+          :name="`${tab}-small`"
```
benjag, line 23: Unique `section` landmarks

mhewson (approved):

PR #131, opened 2023-04-27
Allows the Cedar library to be tree shaken in consuming applications

HeavyMedl:
## Problem

The `cedar.mjs` distributable asset is the total collection of all Cedar library components. Even though `cedar.mjs` is an ES module, Vite/Rollup cannot tree shake the components it needs from `cedar.mjs`. In consuming application builds, this results in Vite/Rollup building the entirety of the Cedar into a module, where it is eventually downloaded by end users.

As a test, I configured Vite to build just the Climber's landing page bundle, stripped all Cedar components from all nested components within the page component. I imported a single Cedar component, `CdrContainer`, to see the resultant bundle.

Screenshot 2023-04-26 at 6 52 17 PM

Within the compiled bundle, I expect to see just the `CdrContainer` code, however, I see the entirety of the Cedar library.

```javascript
// .../target/classes/dist/climber/climber-60b97b4d.js
c({
  name: "CdrAccordion",
  components: { IconCaretDown: Xs },
  props: {
    id: { type: String, required: !0 },
    opened: { type: Boolean, default: !1 },
    compact: { type: Boolean, default: !1 },
    borderAligned: { type: Boolean, default: !1 },
    level: { type: [String, Number], required: !0 },
    contentSpacing: { type: Boolean, default: !0 },
    label: { type: String },
  },
  emits: ["accordion-toggle"],
...
c({
  name: "CdrPopover",
  components: { IconXSm: qr, CdrButton: zr, CdrPopup: rl },
  props: {
    position: {
      type: String,
      required: !1,
      default: "top",
      validator: (e) => re(e, ["top", "bottom", "left", "right"]),
    },
    autoPosition: { type: Boolean, required: !1, default: !0 },
    label: { type: String, required: !1 },
    id: { type: String, required: !0 },
    contentClass: { type: String, required: !1 },
    open: { type: Boolean, default: !1, required: !1 },
  },
  emits: ["opened", "closed"],
...
```

## Solution

This PR sets [preserveModules](https://rollupjs.org/configuration-options/#output-preservemodules) to `true` which breaks up the large `cedar.mjs` distributable into multiple ES modules. This allows the Cedar library to be tree shaken in consuming applications.

I installed the `npm pack`ed tarball from this PR into `climbers-site`. Using the same test setup, we see a major difference at build time:

Screenshot 2023-04-26 at 7 20 05 PM

Now we only see `CdrContainer` code in `climber-3e7f1771.js`.

Thats nearly a **80%** reduction in the uncompressed bundle size (**60%** reduction on the gzipped version). Obviously, this would be a less dramatic difference the more unique Cedar components I used, but its still a major improvement.

## Notes/caveats

I've taken some liberties with this PR that should be examined and reviewed by the team.

### Removed UMD format from Vite config

In order to use `preserveModules`, we need to isolate a ES build in the Vite config. 

```javascript
  build: {
    lib: {
      entry: './src/lib.js',
      formats: ['es'],
      // fileName: (format, entryName) => {
      //   const { base } = path.parse(entryName);
      //   return `${base}.mjs`;
      // },
      fileName: '[name]',
    },
    rollupOptions: options,
  },
```
Using `fileName: '[name]'` preserves the directory structure of the output ES modules within `dist`. This seems like the intuitive thing to do. However, if you prefer a flat output, you can opt to use the commented out function I left.

I'm not sure if anybody uses the UMD module that Cedar outputs, but it case you'd still like this, I've created a separate `vite.umd.config.mjs`. You'd just need to run vite against this in your NPM build script.

### Updated `external` in `rollupOptions.mjs`

I've changed this to use a function that ensures any submodules of these dependencies are externalized. Without this, while using `preserveModules`, submodules from `core-js` and `lodash-es` get built into the distributables, which we don't want.

## History

When we first migrated to Vue 3 and Vite, Vite was on version 2. We validated that tree-shaking was working against apps built with Webpack 4 and Vite 2.

Webpack needed `PURE` annotations in the compiled code in order to tree shake stuff out, while Vite 2 seemed to work out of box.. or so we thought. 

It's unclear whether or not we mistook tree shaking for common chunks generated by Vite/Rollup. Its possible tree shaking hasn't been working since the migration but because we're serving smaller chunks, it appeared tree shaking was happening.

## Reference

- https://github.com/vitejs/vite/issues/8464
- https://github.com/vitejs/vite/issues/5174
- Solution derived from [this comment](https://github.com/vitejs/vite/issues/5174#issuecomment-1193414981).


src/components/accordion/CdrAccordionGroup.vue
```diff
@@ -2,7 +2,7 @@
 import {
   defineComponent, useCssModule, computed, ref, reactive, onMounted, provide,
 } from 'vue';
-import debounce from 'lodash-es/debounce';
+import { debounce } from 'lodash-es';
```
HeavyMedl, line 5: We need to destructure the named exports from the `lodash-es` module, which points to the module entry defined in `lodash-es/package.json`. Otherwise, in the context of testing in consuming applications, Vitest whines about not being able to import `lodash-es/debounce` without a `.js` extension.

vite.umd.config.mjs
HeavyMedl, line 1: Remove this file if UMD is not needed.

rollupOptions.mjs
```diff
@@ -2,11 +2,15 @@ import { babel } from '@rollup/plugin-babel';
...
-  external: ['vue', 'core-js', 'lodash-es', 'tabbable'], // Externalize peerDependencies
+  // external: ['vue', 'core-js', 'lodash-es', 'tabbable'], // Externalize peerDependencies
+  external: (id) => ['vue', 'core-js', 'lodash-es', 'tabbable'].some(
+    (dep) => dep === id || id.startsWith(`${dep}/`),
+  ),
   output: {
     globals: {
       vue: 'Vue',
     },
+    preserveModules: true,
```
HeavyMedl, line 13: The magic

vite.umd.config.mjs
HeavyMedl, line 1: Remove this file if UMD is not needed.

peripateticus (approved):
🎉🎉🎉

vite.config.mjs
```diff
@@ -12,7 +13,12 @@ export default defineConfig({
   build: {
     lib: {
       entry: './src/lib.js',
-      name: 'cedar',
+      formats: ['es'],
+      // fileName: (format, entryName) => {
+      //   const { base } = path.parse(entryName);
+      //   return `${base}.mjs`;
+      // },
```
benjag, line 20: Can this commented code be removed?

benjag (approved):

PR #132, opened 2023-05-02
V13.5.0

benjag:


src/lib.js
```diff
@@ -35,6 +35,5 @@ export { default as CdrToggleButton } from './components/toggleButton/CdrToggleB
 export { default as CdrToggleGroup } from './components/toggleButton/CdrToggleGroup.vue';
 export { default as CdrSwitch } from './components/switch/CdrSwitch.vue'
 export { default as CdrTooltip } from './components/tooltip/CdrTooltip.vue';
-export { default as CdrDocgen } from './component-docgen.json'
```
benjag: Not sure why this was being exported as a JS module when it's already getting copied directly to the dist.

KenjiCrosland: My guess is that this was from before when component-docgen wasn't in the dist.

Just want to make sure the live demo is referencing the proper path.

PR #136, opened 2023-05-09
fix(popup): moves contentClass out of of mapClasses function so it's …

benjag:


src/components/popup/CdrPopup.vue
```diff
@@ -194,7 +194,7 @@ export default defineComponent({
   >
     <div
       v-bind="$attrs"
-      :class="mapClasses(style, 'cdr-popup__content', contentClass)"
+      :class="[style['cdr-popup__content'], contentClass]"
```
benjag, line 197: This was passing the custom `contentClass` prop to the `mapClasses` method which attempts to map them to the style module. The result is that the contentClass was not being applied.

I elected to forgo using the mapClass util as it's only a single class that needs to be modularized for this element.

PR #138, opened 2023-06-06
Feat/typescript

benjag:
WIP Typescript refactor


src/cdr-tokens.d.ts
benjag, line 1: TS requires a definition file for tokens since it doesn't export any types

src/components/input/CdrInput.vue
```diff
@@ -223,8 +226,8 @@ const inputModel = computed({
         :aria-required="required || undefined"
         :aria-invalid="!!error || undefined"
         :aria-errormessage="(!!error && `${uniqueId}-error`) || undefined"
-        v-bind="inputAttrs"
```
benjag, line 243: Removing this because none of the `inputAttrs` are valid for `<textarea>`

src/components/modal/CdrModal.vue
```diff
@@ -329,7 +321,7 @@ export default defineComponent({
         :class="mapClasses(style, 'cdr-modal__contentWrap', 'cdr-modal__dialog')"
         tabIndex="-1"
         :role="role"
-        :aria-modal="!!opened"
+        aria-modal="true"
```
benjag, line 324: This should always be set to true. It was previously, because it was being statically set in the `dialogAttrs` and overwriting this dynamic binding. This change just cleans that up

src/components/accordion/__tests__/__snapshots__/CdrAccordion.spec.js.snap
```diff
@@ -60,7 +60,6 @@ exports[`CdrAccordion > component snapshot when wrapped > renders correctly 1`]
     <button
       aria-controls="test-collapsible"
       aria-expanded="false"
-      aria-pressed="false"
```
benjag, line 63: Redundant when `aria-expanded` is present

PR #141, opened 2023-07-19
CSS code splitting

HeavyMedl:
- Enables CSS code splitting for the library which allows chunks of css to be generated.
- Vite plugin to normalize the name of the CSS and to be output to the expected directory (`dist/style`)
- With this change, `cedar-compiled.css` and `cedar-full.css` are no longer generated. Are they necessary?

This should allow Cedar to deprecate the `extract-css.js` script and remove duplicative dependencies. Should also speed up the build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   "scripts": {
     "prepublishOnly": "npm-run-all lint build",
     "dev": "vite",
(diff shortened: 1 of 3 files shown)


benjag:
Incorporated this into our next release manually (it got a little messy since this was pre-TS). Thanks @HeavyMedl

PR #151, opened 2023-10-27
fix(cdrmodal): fixes the issue where screen readers can read content …

daemoncron:


index.html
```diff
@@ -1,16 +1,46 @@
...
+  <div id="main"></div>
+  <!-- TODO: cover in unit tests, remove this -->
+  <!-- start: testing modal content show/hide -->
+  <div
+    aria-hidden="true"
+    style="opacity: 0;"
+  >something already aria-hidden</div>
+  <div hidden>something display: none;</div>
+  <div style="opacity: 0;">another div of outside content, not aria-hidden</div>
+  <!-- end: testing modal content show/hide -->
```
daemoncron, line 39: TODO this will be removed, for display/testing only

src/components/modal/CdrModal.vue
```diff
@@ -92,11 +102,11 @@ let lastActive: Element | null;
 const modalClosed = ref(!props.opened);
 const isOpening = ref(false);
 
-interface offsetValues {
+interface OffsetValues {
```
daemoncron: Sonarlint caught this formatting.

src/components/modal/CdrModal.vue
```diff
@@ -190,6 +201,62 @@ const removeNoScroll = () => {
   }
 };
```
daemoncron, line 203: Modal scripts begin below

src/components/modal/CdrModal.vue
```diff
@@ -190,6 +201,62 @@ const removeNoScroll = () => {
...
+ */
+const findBackgroundContentNodes = () => {
+  // initial selector is complex, breaking it down
+  const notSelectors = [
+    'body > *',
+    ':not(script)',
+    ':not(style)',
+    ':not([data-is-modal])',
+    ':not([aria-hidden=true])',
+  ]
```
daemoncron, line 220: Summary of these selectors: find all direct children of the body element, that are not:

- script or style tags
- that are not this modal or other modals (they should be hidden when the top modal is open anyway)
- that are not already `aria-hidden`

src/components/modal/CdrModal.vue
```diff
@@ -190,6 +201,62 @@ const removeNoScroll = () => {
...
+  const contentBodyChildren: NodeListOf<HTMLElement> =
+    document.querySelectorAll(notSelectors.join(''));
+
+  // for the remaining, check if they should be hidden
+  contentBodyChildren.forEach((el) => {
+    if (getComputedStyle(el).getPropertyValue('display') !== 'none') {
```
daemoncron, line 226: Also, skip any that have a computed value of `display: none`.

src/components/modal/CdrModal.vue
```diff
@@ -190,6 +201,62 @@ const removeNoScroll = () => {
...
+      // TODO - temp to be sure we tagged right
+      el.setAttribute('data-modal-hide', '');
```
daemoncron, line 228: This isn't strictly necessary, but would be pretty helpful in debugging.  I'm 50/50 on whether to leave it or remove it.

src/components/modal/CdrModal.vue
```diff
@@ -190,6 +201,62 @@ const removeNoScroll = () => {
...
+}
+
+/**
+ * hideModalBackgroundContent
+ * Hide elements tagged onBeforeMount from screen readers
+ */
+const hideModalBackgroundContent = () => {
+  // only run the expensive DOM scraping once
+  if (!backgroundNodesDiscovered) {
+    findBackgroundContentNodes();
```
daemoncron, line 242: Put this here instead of in a `beforeMount` hook, for example.  
Two reasons:

1. Performance: this may never need to run at all.
2. Our pages continue loading content at the body level way past when the `DOMContentLoaded` event fires.  Playing around with this in the PDP, I had to use a setTimeout of something like 5 seconds (on a fast connection) to get all of our little `body` children loaded.

src/components/modal/CdrModal.vue
```diff
@@ -263,21 +333,28 @@ const verticalSpace = computed(() => {
     : windowedSpace + fullscreenSpace;
   // fullscreen, here, would account for outerWrap padding, which is the same CdrSpaceTwoX
 });
-    const scrollMaxHeight = computed(() => totalHeight.value
-    - headerHeight.value
-    - verticalSpace.value);
+
+const scrollMaxHeight = computed(() => totalHeight.value
```
daemoncron, line 337: Moved this out - was throwing a lint error having it inline below.

src/components/modal/CdrModal.vue
```diff
@@ -263,21 +333,28 @@ const verticalSpace = computed(() => {
...
     return 12;
   }
   return 0;
 });
 
+const textContentStyle = computed(() => ({
+  maxHeight: `${scrollMaxHeight.value}px`,
+  paddingRight: `${scrollPadding.value}px`,
+}));
+
```
daemoncron, line 357: Moved this out of template - was throwing a lint error for line length having it inline below.

src/components/modal/CdrModal.vue
```diff
@@ -295,92 +372,91 @@ onUnmounted(() => {
...
 
-
 </script>
 
 <template>
-  <div
-    :class="mapClasses(style, baseClass, !opened ? 'cdr-modal--closed' : '')"
-    ref="wrapperEl"
-  >
+  <Teleport to="body">
```
daemoncron, line 378: template change 1

src/components/modal/CdrModal.vue
```diff
@@ -295,92 +372,91 @@ onUnmounted(() => {
...
     <div
-      :class="[style['cdr-modal__outerWrap'], wrapperClass]"
+      :class="mapClasses(style, baseClass, !opened ? 'cdr-modal--closed' : '')"
+      ref="wrapperEl"
+      data-is-modal
```
daemoncron, line 382: template change 2

src/components/modal/CdrModal.vue
```diff
@@ -295,92 +372,91 @@ onUnmounted(() => {
   window.removeEventListener('resize', handleResize);
 });
 
-
 </script>
 
 <template>
```
daemoncron, line 377: lots of formatting fixes in the modal, noted actual changes below.

src/components/modal/examples/Modal.vue
```diff
@@ -92,7 +92,9 @@ export default {
 },
   data() {
     return {
-      opened: this.$route.name === 'Modals',
+      // TODO not sure why this was setup like this?
+      opened: false,
+      // opened: this.$route.name === 'Modals',
```
daemoncron, line 97: Side fix:  I found this really difficult to develop with - I needed to see the page state before and after modal load. I wasn't sure if there was a reason we had it load on going to the route, but if not I'd love to leave it like this.

src/components/modal/CdrModal.vue
benjag: I'd ask to remove these changes, we use a different config

PR #152, opened 2023-11-02
ACS-3919 - small dev fixes, unit tests

daemoncron:
## Description
**Draft** of some dev changes, before bumping version / adding tags

Mostly formatting and adding some unit tests for modal
No changes to functionality or design, except the `inheritAttrs: false,` which I believe should have always been there (bugfix). 
If someone is relying that bug, it may cause issues, but since adoption is fairly low, we should be able to check individually.
...

### Design:
No design changes

### Cross-browser testing:
NA

### Unit testing:
- [x] Sufficient unit test coverage (see unit test best practices for ideas)
- [x] NA - Snapshot updates are explained with comment and reference the relevant source code change

### A11y:
- [x] Meets WCAG AA standards

### Documentation:

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "url": "https://github.com/rei/rei-cedar.git"
   },
   "scripts": {
+    "prepare": "npm run build",
     "prepublishOnly": "npm-run-all lint build",
     "dev": "vite",
     "build": "vue-tsc && vite build && npm run build:umd && npm run build:extractcss && npm run build:icons && npm run build:docgen",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rei/cedar",
-  "version": "14.1.0",
+  "version": "14.1.1",
   "description": "REI Cedar Component Library",
   "homepage": "https://rei.github.io/rei-cedar/",
   "license": "MIT",
--- a/src/components/modal/CdrModal.vue
+++ b/src/components/modal/CdrModal.vue
@@ -2,10 +2,19 @@
 import { debounce } from '../../utils/debounce'
 import tabbable from 'tabbable';
 import {
-  useCssModule, computed, ref, watch, onMounted, nextTick, onUnmounted, useAttrs,
+  useCssModule,
+  computed,
+  ref,
+  watch,
+  onMounted,
+  nextTick,
+  onUnmounted,
+  useAttrs,
 } from 'vue';
 import {
-  CdrBreakpointSm, CdrSpaceOneX, CdrSpaceTwoX,
+  CdrBreakpointSm,
+  CdrSpaceOneX,
+  CdrSpaceTwoX,
 } from '@rei/cdr-tokens/dist/rei-dot-com/js/cdr-tokens.mjs';
 import onTransitionEnd from './onTransitionEnd';
 import CdrButton from '../button/CdrButton.vue';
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -31,6 +31,7 @@
     "url": "https://github.com/rei/rei-cedar.git"
   },
   "scripts": {
+    "prepare": "npm run build",
```
daemoncron, line 34: To allow for testing off of a git branch

src/components/modal/CdrModal.vue
```diff
@@ -14,14 +23,16 @@ import mapClasses from '../../utils/mapClasses';
 
 /** Disruptive, action-blocking overlays used to display important information */
 defineOptions({
-  name: 'CdrModal'
+  name: 'CdrModal',
+  inheritAttrs: false,
```
daemoncron, line 27: Only actual change, the rest is formatting, just making more friendly for split-screen viewing.

src/components/modal/CdrModal.vue
```diff
@@ -278,6 +290,11 @@ const scrollPadding = computed(() => {
   return 0;
 });
 
+const textContentStyle = computed(() => ({
+  maxHeight: `${scrollMaxHeight.value}px`,
+  paddingRight: `${scrollPadding.value}px`,
+}));
+
```
daemoncron, line 297: moved this logic out of the template.

src/components/modal/__tests__/CdrModal.spec.js
```diff
@@ -35,34 +36,107 @@ describe('CdrModal.vue', () => {
...
-  
+
       wrapper.trigger('keydown', {
         key: 'Escape',
       });
       await wrapper.vm.$nextTick();
-  
+
       expect(wrapper.emitted().closed.length).toBe(2);
     });
```
daemoncron, line 51: Added a few unit tests below for better coverage - I didn't see where we had tests covering the different prop possibilities, let me know if I'm missing something though.

src/components/modal/examples/Modal.vue
```diff
@@ -5,18 +5,20 @@
...
     <cdr-modal
-      label="label text is title or aria"
       :opened="opened"
       @closed="closed"
       wrapper-class="wrapper-test-class"
       overlay-class="overlay-test-class"
       data-backstop="modal"
       role="dialog"
+      data-ui="hamburger-modal"
+      aria-labelledby="some-heading"
```
daemoncron, line 15: Changed this example just because the label & title didn't match.  Also, I wanted to have a version like this to screen reader test.  I did notice the title wasn't reading out on focus with VO with this setup, which is kind of surprising.  May use this to do some later work on focus testing.

src/components/modal/examples/Modal.vue
```diff
@@ -92,7 +94,7 @@ export default {
 },
   data() {
     return {
-      opened: this.$route.name === 'Modals',
+      opened: false,
```
daemoncron, line 97: I hope this is ok. This works much better for testing state before, during, and after modal is launched.

src/components/modal/styles/CdrModal.module.scss
```diff
@@ -17,7 +17,7 @@ $modal-animation-duration: 150ms;
   right: 0;
   top: 0;
   visibility: visible;
-  z-index: 1000;
+  z-index: 9999;
```
daemoncron, line 20: Getting this above gnav so when teleport happens it doesn't end up underneath it.

src/components/modal/examples/Modal.vue
benjag: I believe this was in place for backstop, I don't think it was set to open the modal so it was opened on load.

benjag: Regardless, it's fine since we've stopped using backstop

src/components/modal/CdrModal.vue
benjag: Yep, good catch. This is a bug introduced when we migrated to Vue 3.

benjag (approved):

PR #154, opened 2023-11-07
ACS-3918 - v15 modal teleport

daemoncron:
## Description

- Fixes the issue where screen readers can read content outside the modal in Android Talkback (possibly other combos).
- Upon mounting, any modal instances will now be teleported to the root of the document.body element. When the modal is opened, other top-level children of the body element that could contain screen-readable content will be marked `aria-hidden="true"` to prevent them from being read by certain screen readers.

BREAKING CHANGE: This may not be a breaking change for teams, but they will need to test that it works for their specific implementation, and most importantly, their tests.  Most likely issues will occur around UI tests and and unit tests.  Will have full details in release notes, sending separately to Cedar team.

**Note:** Hide whitespace before you review to clean up template and snapshot diffs.

### Design:
- [x] (NA, no design changes) Reviewed with designer and meets expectations

### Cross-browser testing:
This was covered in internal JIRA ticket, sending link to team separately along with live examples of the issue and fix.
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] iOS
- [x] Android

### Unit testing:
- [x] Sufficient unit test coverage (see unit test best practices for ideas)
- [x] Snapshot updates are explained with comment and reference the relevant source code change

### A11y:
- [x] Meets WCAG AA standards

### Documentation:
- [x] (NA) API docs created/updated
- [x] (NA) Examples created/updated

### DIFF
--- a/src/components/modal/CdrModal.vue
+++ b/src/components/modal/CdrModal.vue
@@ -144,6 +144,7 @@ const handleKeyDown = ({ key }: { key: string }) => {
     default: break;
   }
 };
+
 const handleFocus = (e: Event) => {
   const { documentElement } = document;
   if (modalEl.value?.contains(e.target as HTMLElement) || !documentElement) return;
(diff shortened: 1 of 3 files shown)


src/components/modal/__tests__/__snapshots__/CdrModal.spec.js.snap
daemoncron, line 1: All snapshots just now have `teleport-stub` wrapper.

src/components/modal/CdrModal.vue
```diff
@@ -204,7 +205,62 @@ const addHandlers = () => {
...
+ * Find the nodes that we should aria show/hide
+ * and store them in local state
+ */
+const findBackgroundContentNodes = () => {
+  // initial selector is complex, breaking it down
+  const notSelectors = [
+    'body > *',
+    ':not(script)',
+    ':not(style)',
+    ':not([aria-hidden=true])',
```
daemoncron, line 225: Prevents us from accidentally removing `aria-hidden` from something that was already `aria-hidden="true"` on page load.

benjag (approved):

mhewson (approved):

PR #163, opened 2023-12-13
Fix/lead

benjag:


src/components/lead/CdrHeadingSubheadingBlock.vue
```diff
@@ -19,7 +19,10 @@ const style = useCssModule();
     >
       <slot />
     </CdrHeadingDisplay>
-    <CdrSubheadingSans :class="style[`${baseClass}__subheading`]">
+    <CdrSubheadingSans
```
mhewson, line 22: one question I have is around nesting the subhead span in the header - I think we may want to pass this by accessibility. I like the pattern but am also thinking that the heading may have a char limit for screen readers?

src/components/lead/CdrLead.vue
```diff
@@ -25,7 +25,6 @@ const baseClass = 'cdr-lead'
 
 <template>
   <div :class="style[baseClass]">
-    <div :class="style[`${baseClass}__top`]" />
     <div
       v-if="imgSrc"
```
mhewson, line 29: still question if this should be picture - I guess as a srcset can just be used in image this is fine?

src/components/lead/CdrHeadingSubheadingBlock.vue
benjag: I think it's okay, I used a similar pattern on the cedar doc site and ran it by @daemoncron.

mhewson (approved):

PR #167, opened 2023-12-19
Lead surface

benjag:
Adds surface support to lead container

### DIFF
--- a/src/components/lead/CdrLead.vue
+++ b/src/components/lead/CdrLead.vue
@@ -13,15 +13,15 @@ interface leadProps {
    * @demoSelectMultiple false
    * @values 'top', 'bottom'
    */
-  surfaceOrientation: 'top' | 'bottom'
+  surface?: 'top' | 'bottom'
 }
 
 const props = withDefaults(defineProps<leadProps>(), {
-  surfaceOrientation: 'bottom'
+  surface: 'bottom'
 });
 
 const baseClass = 'cdr-lead'
-const surfaceClass = computed(() => `${baseClass}--${props.surfaceOrientation}`);
+const surfaceClass = computed(() => `${baseClass}--${props.surface}`);
 
 const style = useCssModule();
 </script>
--- a/src/components/lead/CdrLead.vue
+++ b/src/components/lead/CdrLead.vue
@@ -1,24 +1,40 @@
 <script setup lang="ts">
-import { useCssModule } from 'vue';
+import { useCssModule, computed } from 'vue';
+import mapClasses from '../../utils/mapClasses';
 
 /** Container used for assembling leads */
 defineOptions({
   name: 'CdrLead'
 });
 
-const style = useCssModule();
+interface leadProps {
+  /** 
+   * Sets the orientation of the background surface
+   * @demoSelectMultiple false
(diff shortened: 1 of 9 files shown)


src/components/lead/CdrLead.vue
```diff
@@ -1,24 +1,40 @@
...
 });
 
-const style = useCssModule();
+interface leadProps {
+  /** 
+   * Sets the orientation of the background surface
+   * @demoSelectMultiple false
+   * @values 'top', 'bottom'
+   */
+  surfaceOrientation: 'top' | 'bottom'
```
mhewson: surfaceOrientation seems a bit of a mouthful would surface:top /bottom do the same?

src/components/lead/examples/Lead.vue
```diff
@@ -10,14 +10,24 @@
...
+      <template #top>
         <CdrImg src="/src/dev/static/cedar-1920x1080.jpg" />
       </template>
-      <template #content>
+      <template #bottom>
         <CdrTitle>Global lead title</CdrTitle>
       </template>
     </CdrLead>
 
+    <h2>Lead container with surface on top</h2>
```
mhewson, line 21: nice!

src/components/lead/styles/vars/CdrLead.vars.scss
```diff
@@ -1,10 +1,18 @@
 @mixin cdr-lead-base-mixin() {
   width: 100%;
   line-height: 0;
-  background: linear-gradient(
-    transparent 0 50%,
-    var(--cdr-lead-base-background-color, var(--cdr-color-background-primary, #{$cdr-color-background-primary})) 26% 100%,
-  );
+  &--top {
+    background: linear-gradient(
+      var(--cdr-lead-base-background-color, var(--cdr-color-background-primary, #{$cdr-color-background-primary})) 0 50%,
```
mhewson: is this the surface color - the background color should come from leads parent

mhewson (approved):
looks grand!

src/components/lead/styles/vars/CdrLead.vars.scss
benjag: Meaning the custom prop should be named `--cdr-lead-surface-color`? The tokens for surface don't exist so the fallbacks would still be background color right?

src/components/lead/styles/vars/CdrLead.vars.scss
```diff
@@ -3,14 +3,14 @@
   line-height: 0;
   &--top {
     background: linear-gradient(
-      var(--cdr-lead-base-background-color, var(--cdr-color-background-primary, #{$cdr-color-background-primary})) 0 50%,
+      var(--cdr-lead-surface-color, var(--cdr-color-background-primary, #{$cdr-color-background-primary})) 0 50%,
```
mhewson, line 6: I wonder if there is time to get 
--cdr-surface
--cdr-surface-recessed 
into tokens and then default the surface color to --cdr-surface - and allow the prop to switch its token?

benjag: Can follow up with that, I want to get started on a round of integration testing

PR #170, opened 2024-01-04
Dependency maintenance

benjag:
This PR brings in updates to Node, Vite, Vitest along with other dependencies.

Updating vitest required a number of changes to the library:
- Deleted `vitest.config.ts` and moved test configuration to `vite.config.ts` (per vitest recommendation)
- Changed how tests handled style modules, previously it was processing them but returning them un-hashed
  -  Now, the tests skip module processing altogether, as a result a number of snapshots needed to be updated
  - Additionally, the broken snapshots uncovered an opportunity to clean up size prop usage in the library. Not all components were using the shared prop util when able to, and some were using it when they shouldn't, leading the component to expose an invalid size option (wasn't breaking, the prop option just didn't do anything)


vite.config.ts
```diff
@@ -37,6 +38,16 @@ export default defineConfig({
       '~': fileURLToPath(new URL('./node_modules', import.meta.url)),
     },
   },
+  test: {
+    globals: true,
+    exclude: [...configDefaults.exclude, '**/test/e2e', '**/templates/__tests__'],
+    environment: 'jsdom',
+    css: {
+      modules: {
+        classNameStrategy: 'non-scoped'
```
benjag, line 47: See https://vitest.dev/config/#css-modules-classnamestrategy

Note: The warning on Vitest now bypassing CSS module processing

src/components/switch/__tests__/__snapshots__/Switch.spec.js.snap
```diff
@@ -25,7 +25,7 @@ exports[`CdrSwitch > matches snapshot 1`] = `
   >
     <svg
       aria-hidden="true"
-      class="cdr-icon cdr-icon--medium cdr-switch__button-icon"
+      class="cdr-icon cdr-switch__button-icon"
```
benjag, line 28: An example where an invalid size class was being applied only in the test environments

src/components/rating/components/CdrStar00.vue
```diff
@@ -1,17 +1,10 @@
...
 const props = defineProps({
-  size: {
-      type: String,
-      default: 'medium',
-      validator: (value: string) => validateProp(
-        value,
-        ['small', 'medium', 'large'],
-      ),
-    },
+  size: sizeProps,
```
benjag, line 7: An example of a component that was able to use the common `size` prop util but wasn't

src/components/select/CdrSelect.vue
```diff
@@ -58,11 +57,11 @@ const props = defineProps({
...
   /**
-   * Sets the component's size; values can target responsive breakpoints. Example `small@lg`
-   * @demoSelectMultiple false
-   * @values small, medium, large
+   * Sets the component's size; values can target responsive breakpoints. Example `large@lg`
+   * @demoSelectMultiple true
+   * @values large
   */
-  size: sizeProps,
+  size: String,
```
benjag, line 64: An example of where a component was using the common `size` prop util but should't have been because it was introducing invalid options.

KenjiCrosland (approved):

mhewson (approved):

PR #188, opened 2024-11-02
Mark `@rei/cdr-tokens` as external

HeavyMedl:
### Problem

I see an SSR runtime exception from Nitro (Nuxt) derived from dead references in `@rei/cedar`'s distributed modules. When Nuxt builds the application (`nuxt build`), it outputs optimized server assets to `.output/server,` where you'd run a productionized Nuxt app from. 

The Nuxt build heuristics exclude nested `node_modules` directories, instead hoisting them to the top-level `.output/server/node_modules`. However, some `@rei/cedar` modules are dependent on those modules existing in a relative location:

```typescript
// .output/server/node_modules/@rei/cedar/dist/src/components/image/CdrImg.vue2.mjs
import { CdrRadiusSoft as f, CdrRadiusSofter as C, CdrRadiusRound as y } from "../../../node_modules/@rei/cdr-tokens/dist/rei-dot-com/js/cdr-tokens.mjs";
```

This causes the following Nuxt/SSR runtime exception:

```
[Vue Router warn]: uncaught error during route navigation:
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/kmedley/code/alpine-composer/.output/server/node_modules/@rei/cedar/dist/node_modules/@rei/cdr-tokens/dist/rei-dot-com/js/cdr-tokens.mjs' imported from /Users/kmedley/code/alpine-composer/.output/server/node_modules/@rei/cedar/dist/src/components/modal/CdrModal.vue2.mjs
    at finalizeResolution (node:internal/modules/esm/resolve:265:11)
    at moduleResolve (node:internal/modules/esm/resolve:933:10)
    at defaultResolve (node:internal/modules/esm/resolve:1157:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:390:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:359:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:234:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:87:39)
    at link (node:internal/modules/esm/module_job:86:36) {
  code: 'ERR_MODULE_NOT_FOUND',
  url: 'file:///Users/kmedley/code/alpine-composer/.output/server/node_modules/@rei/cedar/dist/node_modules/@rei/cdr-tokens/dist/rei-dot-com/js/cdr-tokens.mjs'
}
```
In the context of the Alpine ecosystem, this hasn't been a problem because our processes use a customized SSR setup that copies `node_modules` into the Crampon SSR sidecar (docker container). 

### Solution

This PR marks `@rei/cdr-tokens` as external, so the build system doesn't bundle them into the `@rei/cedar` distributed modules. Instead, Cedar modules will directly reference `@rei/cdr-tokens` from `node_modules` at run-time, which should be the expected behavior.

### Side-effects

With `@rei/cdr-tokens` externalized, we need to adjust module entry points for `@rei/cedar`, as Vite no longer outputs `.dist/src`. Those entry points will live in `.dist/`, thus, we update references in `package.json`. Additionally, module references to `@rei/cdr-tokens` are updated to match.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ src/**/styles/themes/*
 
 src/**/package-lock\.json
 unmigrated/
+
+## Dev output for hosting
+pages
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,12 +9,12 @@
       "version": "15.2.1",
       "license": "MIT",
       "dependencies": {
+        "@rei/cdr-tokens": "^12.3.0",
         "tabbable": "^4.0.0"
       },
       "devDependencies": {
         "@babel/preset-env": "^7.18.2",
         "@emotion/is-prop-valid": "^1.2.1",
-        "@rei/cdr-tokens": "^12.3.0",
         "@rei/cedar-icons": "^2.6.1",
         "@rollup/plugin-babel": "^6.0.3",
         "@types/tabbable": "^3.1.2",
@@ -3067,7 +3067,6 @@
     "node_modules/@rei/cdr-tokens": {
       "version": "12.3.0",
       "resolved": "git+ssh://git@github.com/rei/rei-cedar-tokens.git#02446ea636a23bd896c7fadcb58e60c80d1ff709",
-      "dev": true,
       "license": "ISC",
       "engines": {
         "node": ">= 20.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@rei/cedar",
-  "version": "15.2.0",
+  "version": "15.2.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@rei/cedar",
-      "version": "15.2.0",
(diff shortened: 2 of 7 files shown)


mhewson (approved):

HeavyMedl (commented):
Looking at how `@rei/cdr-tokens` is used, e.g., the JS API for some components uses it directly, means it’s actually a dependency, not a devDep. We want to ensure the lib gets the version of tokens it needs now that it’s (correctly) being treated as an external module.

sikhote (approved):
Checked within our doc site for kicks and things worked as expected. Not sure why this was not the original method for importing tokens, but this makes sense.

sikhote:
@HeavyMedl & @mhewson I can merge/release—any objections or concerns?

HeavyMedl:
Ship it 🚢

PR #190, opened 2024-11-13
Pr/membership

mhewson:
## Description
membership palettes


c4rlosviteri (approved):

src/styles/main.scss
```diff
@@ -1,5 +1,6 @@
 /* Tokens */
 @import '/node_modules/@rei/cdr-tokens/dist/rei-dot-com/scss/cdr-tokens';
+@import '/node_modules/@rei/cdr-tokens/dist/rei-dot-com/css/cdr-tokens';
```
sikhote: Do we need both of these? (line 2 & 3)

src/components/surface/CdrSurface.vue
```diff
@@ -18,20 +18,25 @@ const props = withDefaults(defineProps<surface>(), {
...
+    class: mapClasses(style, ...classes) || undefined,
+  };
 });
 </script>
 
 <template>
   <component
     :is="tag"
     v-bind="rootProps"
+    :data-palette="props.palette"
```
sikhote: I think this is better added to the `rootProps` return, something like 

```
  return {
    class: mapClasses(style, ...classes) || undefined,
    'data-palette': props.palette,
  };
```

src/components/surface/examples/Surface.vue
```diff
@@ -36,20 +36,15 @@ const boxes: Example[] = [
   {
     label: 'This surface uses the primary modifier to apply a background and border color',
```
sikhote: I suggest this label is updated to reflect the example, maybe `Showing the use of `membership-subtle` palette

sikhote: Same with the label below as well

src/dev/SinkWrapper.vue
```diff
@@ -37,7 +37,7 @@ export default {
   data() {
     return {
       palette: this.$route.query.palette || 'default',
-      palettes: ['default', 'sandstone'],
+      palettes: ['default', 'sandstone', 'membership-vibrant', 'membership-subtle'],
```
sikhote, line 40: 👍

src/styles/cdr-reset.scss
```diff
@@ -114,9 +114,4 @@ input[type='month'] {
   -webkit-appearance: none;
 }
 
-[data-palette='sandstone'] {
-  --cdr-color-background-surface-primary: #{$cdr-color-background-secondary};
-  --cdr-color-background-surface-secondary: #dcd6cb;
-  --cdr-color-border-surface-primary: #{$cdr-color-border-secondary};
-  --cdr-color-border-surface-secondary: #726d64;
-}
+
```
sikhote: With this removal, how are users expected to add in palettes? Maybe they would need to import the palettes file manually?

sikhote:
Overall looks good, just looking to clarify a few things

src/styles/settings/palette.vars.scss
```diff
@@ -0,0 +1,30 @@
+
+[data-palette='sandstone'] {
+  --cdr-color-background-surface-primary: #{$cdr-color-background-secondary};
+  --cdr-color-background-surface-secondary: #dcd6cb;
+  --cdr-color-border-surface-primary: #{$cdr-color-border-secondary};
+  --cdr-color-border-surface-secondary: #726d64;
+}
+
+[data-palette='membership-subtle'] {
+  --cdr-color-background-surface-primary: var(--cdr-membership-subtle-color-background-primary);
```
sikhote: If these get changed to sass variables, solves the double import issue perhaps

src/components/surface/CdrSurface.vue
```diff
@@ -9,28 +9,72 @@ import { getSurfaceProps } from '../../utils/surface';
 defineOptions({ name: 'CdrSurface' });
 
 const props = withDefaults(defineProps<surface>(), {
+    /**
```
sikhote: FYI this type of documentation is also part of the `surface` interface type. I'm not sure which one the doc site will pick up. I could see it as a little confusing to have both for someone coming into it without background knowledge, but I don't mind.

src/styles/cdr-reset.scss
sikhote: I think this question still applies—should we export the palette sass file you created?

c4rlosviteri (approved):

src/components/surface/CdrSurface.vue
mhewson: Oh my bad - interface seems like the correct location for this - the doc site seems to be using slightly different language

src/styles/cdr-reset.scss
mhewson: I ended up moving this file to styles root dir and allowing it to be built to dist

src/styles/settings/palette.vars.scss
mhewson: yes I added sass as a fallback

sikhote (approved):
Sweeet ✅

PR #192, opened 2024-12-02
Media Object

sikhote:
## Description
- Creates MediaObject component using Layout component
- Lots of examples
- Some new supporting Sass mixins
- Some new supporting JS utilities
- Updated and pinned vue-tsc and typescript because of https://github.com/vuejs/language-tools/issues/5018

https://rei.github.io/rei-cedar/


sikhote:
Added in some unit tests.

sikhote:
Sorry for the noise, this is ready for final review now.

src/components/mediaObject/CdrMediaObject.vue
```diff
@@ -0,0 +1,119 @@
...
+/** Component for buttons that have a checked state. */
+
+defineOptions({ name: 'CdrMediaObject' });
+
+const props = withDefaults(defineProps<MediaObject>(), {
+  align: 'start',
+  mediaPosition: 'left',
+  mediaWidth: '1fr',
+  mediaHeight: 'auto',
+  cover: false,
```
mhewson: I know this is mostly understood as an image property but as we are using mediaX do we want this to follow that pattern? 
as in all props relating to the media get the media prefix and any related to content get content?

src/components/mediaObject/CdrMediaObject.vue
```diff
@@ -0,0 +1,119 @@
+<script setup lang="ts">
+import { useCssModule, computed } from 'vue';
+import mapClasses from '../../utils/mapClasses';
+import { MediaObject, NameValuePair, HtmlAttributes } from '../../types/interfaces';
+import type { Breakpoint } from '../../types/other';
+import { modifyClassName } from '../../utils/buildClass';
+import { getLayoutStyling } from '../../utils/mediaObject';
+import CdrLayout from '../layout/CdrLayout.vue';
+import { breakpoints, spacing } from '../../utils/other';
```
mhewson, line 9: rad

src/components/mediaObject/CdrMediaObject.vue
```diff
@@ -0,0 +1,119 @@
...
+
+      breakpoints.forEach((breakpoint: Breakpoint) => {
+        // Add in padding styles for various breakpoints
+        inlineStyles[`--cdr-media-object-content-padding-${breakpoint}`] =
+          spacing[contentPadding[breakpoint]];
+      });
+    }
+  }
+
+  // Add overlay class, which skips most other props
```
mhewson: what does skips most other props mean? invalidates? ignores?

src/components/mediaObject/CdrMediaObject.vue
```diff
@@ -0,0 +1,119 @@
...
+    class: mapClasses(style, ...classes) || undefined,
+    style: inlineStyles,
+  };
+});
+</script>
+
+<template>
+  <CdrLayout v-bind="rootProps">
+    <div :class="style['cdr-media-object__media']">
+      <!-- @slot Where the media should be placed. Should be a single node. -->
```
mhewson, line 110: in the instance of a Position object, or several, is the expectation that they would be in their own slot or that they would share a slot with media?

src/components/mediaObject/examples/MediaObject.vue
```diff
@@ -0,0 +1,336 @@
+<script setup lang="ts">
```
mhewson, line 1: An example I would like to capture

full screen -- lead layout with gap between image and content block
mid screen more of the split surface component idea -- horizontal layout 
small screen -- stack as if in card

mhewson (approved):

src/components/mediaObject/CdrMediaObject.vue
sikhote: Hmm, we may need to chat about this one. If multiple images are passed (in the future), it would change up how this slot functions.

src/components/mediaObject/examples/MediaObject.vue
sikhote: I am adding in a `gap` to the `pass down props to Layout and Surface` example right now. 

Lets chat about the others, I'm not sure what those examples would show that some of the existing samples are not already showing.

src/components/mediaObject/CdrMediaObject.vue
sikhote: Updated the comment, but to answer: a lot of the props are just not used, so they are ignored when in overlay

src/components/mediaObject/CdrMediaObject.vue
sikhote: Updated to `mediaCover` for consistency

src/components/mediaObject/CdrMediaObject.vue
mhewson: yes I can see the misuses pilling up on product side - so how to nail this down... docs?

src/components/mediaObject/examples/MediaObject.vue
mhewson: maybe that example could also use srcset?

src/components/mediaObject/CdrMediaObject.vue
sikhote: I will create an example with badge inside of image

sikhote: Added in a badge example to kitchen sink

src/components/mediaObject/examples/MediaObject.vue
sikhote: Agreed that this will be part of showcase or doc site in some way

PR #193, opened 2024-12-09
testing e2e

sikhote:
## Description

- Moved all e2e testing to Playwright, moving away from Nightwatch
- No longer needed the e2e page, just using the kitchen sink page

### Why
Tests were failing with Nightwatch and I could not find any support as to why. It worked last Friday. Playwright seems much more supported and the current standard.


c4rlosviteri (approved):

mhewson (approved):
Ok, I do like the interface for generating the playwright report tho I did create a reports dir for the linting work - I wonder if we can store these together to ensure we capture a full backlog of regressions that need to be managed?

sikhote:
I think the linting work will be much more important. For the playwright reports, to me they are "throw away", because any issues they catch should be fixed before merging a PR. Any reports they produce are not really helpful until there is an issue...but maybe that's just me 🙃

PR #197, opened 2024-12-26
media-object-refinement

sikhote:
## Description

- CdrChoreographer: updates to support more components
- CdrLayout: always has `container-type: inline-size;` applied
- CdrLayout: now accepts a `containerName` prop
- CdrMediaObject: overlay works slightly differently, applying a media height and not relying on media's natural height
- CdrMediaObject: applies media and content styles differently after use in other apps
- CdrMediaObject: expanded examples
- CdrMediaObject: changed some structure of sass mixins for easier reuse
- CdrMediaObject: inner img, picture, video are applied `display: block`
- CdrMediaObject" inner content has `container-type: inline-size;` applied for fluid content compatibility
- interfaces: updates for accuracy

https://rei.github.io/rei-cedar/


src/components/mediaObject/__tests__/__snapshots__/CdrMediaObject.spec.js.snap
```diff
@@ -354,21 +302,17 @@ exports[`CdrMediaObject > snapshot tests > media width dynamic 1`] = `
 exports[`CdrMediaObject > snapshot tests > overlay, row align, column align, content configured independently to 50% width 1`] = `
 <div
   class="cdr-layout cdr-layout--rows cdr-layout--columns cdr-media-object cdr-media-object--overlay"
-  style="--cdr-layout-rows: auto; --cdr-layout-columns: auto; --cdr-media-object-row-align: center; --cdr-media-object-column-align: end;"
+  style="--cdr-layout-rows: 56cqw; --cdr-layout-columns: auto; --cdr-media-object-row-align: center; --cdr-media-object-column-align: end;"
```
c4rlosviteri, line 305: Did you test this on Firefox? https://caniuse.com/mdn-api_css_cqw_static

sikhote: I think this is the link of importance:
https://caniuse.com/mdn-css_types_length_container_query_length_units

With that said, you do bring up a good point about browser support. If we look at the updated browser matrix:
https://rei-coop.atlassian.net/wiki/spaces/FED/pages/229213181/REI+Supported+Browser+Matrix

It looks like iOS safari still does not support container queries (it is supported on v16). I think we should have a chat about all this.

c4rlosviteri: My bad, you are right! And we should introduce a polyfill for now. Thoughts @mhewson?

sikhote: Something like this? https://github.com/GoogleChromeLabs/container-query-polyfill

Sounds helpful

c4rlosviteri: Yes, that one should be enough for polyfilling iOS 15

c4rlosviteri (approved):

mhewson (approved):

PR #201, opened 2025-01-31
Pr/surface navigation

c4rlosviteri:
# Surface Navigation
## Overview
This PR introduces the Surface Navigation component, a specialized variant of Cedar's Surface component designed specifically for navigational elements. It provides comprehensive state management and styling capabilities for interactive surfaces.

## Key Features
- State-based styling for rest, hover, active, and visited states
- Configurable background colors, borders, shadows, and radii
- Accessibility-focused implementation with proper focus states
- TypeScript support with strict typing
- Modular SCSS architecture with mixins and variables

## Testing
- Component includes examples demonstrating various configurations
- Tested across all supported browsers
- Verified keyboard navigation and focus states
- Confirmed WCAG compliance for all states

## Documentation
- Added comprehensive documentation with usage examples
- Included TypeScript interfaces and prop definitions
- Documented accessibility considerations and best practices

## Usage Example
```vue
<CdrSurfaceNavigation 
  :background="{
    rest: 'primary',
    hover: 'secondary'
  }"
  :withBorder="{
    rest: false,
    hover: true
  }"
>
  <CdrSurfaceNavigationLink>
    Navigation Content
  </CdrSurfaceNavigationLink>
</CdrSurfaceNavigation>

### CODE REVIEW

File: src/components/surfaceNavigation/examples/SurfaceNavigation.vue [hunk 1]
```diff
@@ -0,0 +1,162 @@
...
+const handleFocus = (e: FocusEvent) => {
+  console.log('focused', e);
+}
+</script>
+
+<template>
+  <div class="example">
+    <h2>Surface Navigation</h2>
+    <template
+      v-for="{ exampleName, label, title, image, props, ratingProps, price, mediaObjectProps } in boxes"
```
mhewson, line 85: I would be great to see a few more examples of this with random properties set, use this as our edge case test -- 

I would expect that for instance surface navigation without a media object - similar to fulfillment tile
nested surface navigations for some horrible situation gone wrong and so on

mhewson: the updates are looking great, however I still would like to see enough examples to get a good sense of our edge cases

src/components/fulfillmentTile/styles/CdrFulfillmentTileIcon.module.scss
```diff
@@ -1,26 +1,68 @@
-@import '../../../styles/settings/index';
-@import './vars/CdrFulfillmentTileIcon.vars';
-
 .cdr-fulfillment-tile-icon {
```
mhewson, line 1: this looks great, as a nit pick I would like to see these property value pairs be alphabetized

src/components/surface/CdrSurface.vue
```diff
@@ -1,47 +1,30 @@
...
 
 const props = withDefaults(defineProps<surface>(), {
-  background: 'primary',
-  borderColor: 'primary',
-  borderStyle: 'solid',
-  borderWidth: 'sixteenth-x',
-  modifier: 'default',
-  radius: 'sharp',
-  shadow: 'flat',
+  variant: 'base',
```
mhewson: is 'base' more appropriate here than 'default'?

src/components/surface/examples/Surface.vue
```diff
@@ -26,7 +26,7 @@ const boxes: Example[] = [
     props: {
       radius: 'soft',
       background: 'secondary',
-      shadow: 'floating',
+      boxShadow: 'floating',
```
mhewson, line 29: I think these should be named after the design foundations they implement rather than the css prop we use to expose them  - I think this one is "prominence"

src/components/surface/styles/CdrSurface.module.scss
```diff
@@ -1,10 +1,198 @@
-@import '../../../styles/settings/index';
-@import './vars/CdrSurface.vars';
+@use "sass:map";
+
+// Property maps
+$properties: (
+  background: (
+    primary: var(--cdr-color-background-primary),
+    secondary: var(--cdr-color-background-secondary),
+    brand-spruce: var(--cdr-color-background-brand-spruce),
```
mhewson: lets rename this to just brand - brand-spruce seems really limiting from the context of what would happen via a palette being used

mhewson: which may change the way I go about setting the palettes up actually - rather than the primary color being mapped to a new value I may consider just updating the brand color for things like the membership palette

src/components/surface/styles/CdrSurface.module.scss
```diff
@@ -1,10 +1,198 @@
...
+    sale: var(--cdr-color-background-sale)
+  ),
+  border-color: (
+    primary: var(--cdr-color-border-primary),
```
mhewson, line 13: anyway to do something like
allow the user to set a context, like primary, secondary, error - and then not have to define each prop unless its different than the generic context? 

so something like 

surface-secondary (this sets normal secondary values for background, border color, ect) and then
user also defines 
radius: round  or some such thing

src/components/surface/styles/CdrSurface.module.scss
```diff
@@ -1,10 +1,198 @@
...
+    raised: var(--cdr-prominence-raised),
+    elevated: var(--cdr-prominence-elevated),
+    floating: var(--cdr-prominence-floating),
+    lifted: var(--cdr-prominence-lifted)
+  )
+);
+
+$states: (rest, hover, active, checked, loading, disabled, visited);
+$css-vars: (
+  background: --background,
```
mhewson: I wonder if we want the local variables to be something like --cdr-surface-background -- this way if someone ends up defining something like --background at their root scope we would not have variable naming collisions - additionally as these variables are still available to view in rendered code we have an easy signature "--cdr" to search for to see things like our overall usage

src/components/surface/styles/CdrSurface.module.scss
```diff
@@ -1,10 +1,198 @@
...
-    $surface-border-colors,
-    $surface-component
-  );
+  appearance: none;
+  display: block;
+  position: relative;
+  text-align: left;
+
+  // Base CSS custom properties
+  --background: var(--cdr-surface-background, var(--cdr-surface-background-rest));
```
mhewson: we will want to add the sass vars in as a fallback for our sass only token importers -- or come up with a way to ensure they have that asset available

src/components/surface/styles/CdrSurface.module.scss
```diff
@@ -1,10 +1,198 @@
...
+  // Generate base styles
+  @each $prop, $values in $properties {
+    @each $value-name, $value in $values {
+      &[data-#{$prop}='#{$value-name}'] {
+        #{map.get($css-vars, $prop)}: #{$value};
+      }
+    }
+  }
+
+  // Generate state-based styles
```
mhewson, line 93: this seems a bit fiddly 

how about something like 

@each $state in $states {
  @if $state != 'rest' {
    $selector: if($state == 'hover', 
                  '&:hover:not([aria-pressed="true"], [aria-checked="true"], :disabled)',
                  if($state == 'active', 
                     '&:active, &:focus',
                     '&[data-#{$state}="true"]'));
    
    @each $prop, $values in $properties {
      @each $value-name, $value in $values {
        &[data-#{$prop}-#{$state}='#{$value-name}'] {
          #{$selector} {
            #{map.get($css-vars, $prop)}: #{$value};
          }
        }
      }
    }
  }
}

src/components/surface/styles/CdrSurface.module.scss
```diff
@@ -1,10 +1,198 @@
-@import '../../../styles/settings/index';
-@import './vars/CdrSurface.vars';
+@use "sass:map";
+
+// Property maps
```
mhewson, line 3: lets add a TODO comment to get this from cdr-tokens

src/components/surface/styles/CdrSurface.module.scss
```diff
@@ -1,10 +1,198 @@
...
+  border-radius: var(--border-radius);
+  box-shadow: var(--box-shadow);
+  
+  &::before {
+    border: var(--border-width) var(--border-style) var(--border-color);
+    border-radius: inherit;
+    content: '';
+    inset: 0;
+    position: absolute;
+    transition: border-color var(--cdr-duration-3-x) var(--cdr-timing-function-ease);
```
mhewson, line 124: we should also use the will-change prop for selectors that will animate

src/components/surfaceNavigation/CdrSurfaceNavigationLink.vue
```diff
@@ -0,0 +1,48 @@
+<script setup lang="ts">
```
mhewson: should this use cdr-link as the base? if not it probably also needs to support 
:target="target"
:rel="computedRel"

src/utils/mapClasses.ts
```diff
@@ -1,8 +1,18 @@
+/**
```
mhewson, line 1: nice! I appreciate the added docs

src/types/interfaces.ts
```diff
@@ -2,12 +2,12 @@ import type { Component } from 'vue';
 import type {
   Tag,
   Space,
+  SpaceFixed,
```
mhewson, line 5: this is static rather than fluid space?

src/types/interfaces.ts
```diff
@@ -2,12 +2,12 @@ import type { Component } from 'vue';
...
   SpaceOption,
   Shadow,
   Radius,
   BorderColor,
   BorderStyle,
-  Modifier,
```
mhewson, line 10: yes that was an oddball

src/types/interfaces.ts
```diff
@@ -230,101 +230,85 @@ export interface Layout extends NameValuePair {
   containerName?: string;
 }
 
+interface StateConfig<T> {
```
mhewson: Where do you think we can capture out TS patterns and best practices like noting that we use T as a placeholder for another type that will be specified when StateConfig is used.  or are some things just a generic convention?

mhewson (approved):

src/components/fulfillmentTile/styles/CdrFulfillmentTileIcon.module.scss
```diff
@@ -1,14 +1,14 @@
...
-  width: 2.2rem;
-  height: 2.2rem;
-  background: white;
-  display: flex;
   align-items: center;
-  justify-content: center;
+  background: white;
```
mhewson: should be --cdr-color-background-primary?

src/components/fulfillmentTile/styles/CdrFulfillmentTileIcon.module.scss
```diff
@@ -1,14 +1,14 @@
...
   border-radius: var(--cdr-radius-round);
-  position: relative;
+  display: flex;
   fill: var(--cdr-fulfillment-tile-icon-fill);
   flex-shrink: 0;
+  height: 2.2rem;
```
mhewson, line 8: Note for future -- We have Icon spacing implemented in several components - I want to centralize this logic so the icon sizes stay consistent -- look at banner and icon button as examples

src/components/fulfillmentTile/styles/CdrFulfillmentTileIcon.module.scss
```diff
@@ -1,14 +1,14 @@
...
+  justify-content: center;
+  position: relative;
+  width: 2.2rem;
```
mhewson, line 11: also token

src/components/fulfillmentTile/styles/CdrFulfillmentTileIcon.module.scss
mhewson: or data-background="primary"

PR #206, opened 2025-02-28
Add CdrObjectOverlay Component with Positioning and Customization Options

c4rlosviteri:
This pull request introduces the CdrObjectOverlay component—a flexible overlay solution for positioning content relative to a container in nine different positions. The component is designed to be highly customizable, allowing developers to adjust the positioning, margin, and even the HTML tag used for the overlay container.

# Key Features
- Flexible Positioning: The component accepts a position prop that supports the following values:

  - top-left, top-center, top-right
  - left, center, right
  - bottom-left, bottom-center, bottom-right
  - The default position is set to center.

- Customizable Margin: The margin prop allows developers to specify the spacing around the overlay content. By default, this is set to 0.

- HTML Tag Customization: With the tag prop, you can define the HTML tag for the container element. The default is div, but it can be changed to any valid HTML tag (e.g., section, article).

- Slots for Content: The component defines two slots:

  - container: For the content the overlay will be positioned relative to.
  - content: For the actual overlay content.

This PR adds a versatile overlay component that aligns with our existing design system and development patterns. The tests ensure reliability and consistency, preparing this component for integration into larger projects.


src/components/objectOverlay/CdrObjectOverlay.vue
```diff
@@ -0,0 +1,143 @@
+<script setup lang="ts">
+import { useCssModule, computed } from 'vue';
+import {
+  ObjectPosition,
+  ResponsivePosition,
+  ResponsiveSpace,
+  Spacing,
+  Gradient,
+  ResponsiveGradient
+} from '../../types/interfaces';
```
mhewson: at what point do we want to start types/interfaces/"file" or "pattern"

src/components/objectOverlay/styles/CdrObjectOverlay.module.scss
```diff
@@ -0,0 +1,132 @@
...
+  'center-center': (top: 50%, left: 50%, transform: translate(-50%, -50%)),
+  'right-center': (top: 50%, right: 0, transform: translateY(-50%)),
+  'left-bottom': (bottom: 0, left: 0, transform: none),
+  'center-bottom': (bottom: 0, left: 50%, transform: translateX(-50%)),
+  'right-bottom': (bottom: 0, right: 0, transform: none)
+);
+
+// Dark theme gradients (default)
+$dark-gradients: (
+  'to-top': linear-gradient(to top, black 0%, rgba(0, 0, 0, 0.5) 20%, transparent 40%),
```
mhewson, line 25: Is that the updated black value from Michelle?

mhewson: also I think these are good candidates for custom props, and gradient tokens

mhewson:
Screenshot 2025-03-18 at 7 54 16 AM
It looks like the gradient is clipping, but the overlay content is not

mhewson (approved):

PR #231, opened 2025-11-20
Cdr-Input error "live" region 

saagude:
## Description

Live region exists even if error is not present, so when the error is displayed SR will announce it. 
https://rei-coop.atlassian.net/browse/CDR-3489

### Design:

### Cross-browser testing:
- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] iOS
- [ ] Android

### Unit testing:

### A11y:
- [ ] Meets WCAG AA standards

### Documentation:

### DIFF
--- a/src/components/formError/__tests__/__snapshots__/CdrFormError.spec.js.snap
+++ b/src/components/formError/__tests__/__snapshots__/CdrFormError.spec.js.snap
@@ -4,8 +4,8 @@ exports[`CdrFormError > using an error slot > renders correctly 1`] = `
 <div
   class="cdr-form-error --active-error"
 >
-  <div
-    class="cdr-form-error__icon cdr-form-error__icon--active"
+  <span
+    class="cdr-form-error__icon"
   >
     <svg
       aria-hidden="true"
@@ -30,7 +30,7 @@ exports[`CdrFormError > using an error slot > renders correctly 1`] = `
       
       <!--v-if-->
     </svg>
-  </div>
+  </span>
   <div
     aria-atomic="true"
     aria-relevant="all"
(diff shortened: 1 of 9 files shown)


src/components/formError/__tests__/__snapshots__/CdrFormError.spec.js.snap
```diff
@@ -2,11 +2,10 @@
 
 exports[`CdrFormError > using an error slot > renders correctly 1`] = `
 <div
-  class="cdr-form-error"
-  role="status"
+  class="cdr-form-error --active-error"
 >
-  <span
-    class="cdr-form-error__icon"
+  <div
```
cviterirei: Please just make sure that when you change from span to div the styles don't break. span is inline and div block by default. If you already reviewed this ignore this comment.

src/components/formError/__tests__/__snapshots__/CdrFormError.spec.js.snap
```diff
@@ -31,20 +30,28 @@ exports[`CdrFormError > using an error slot > renders correctly 1`] = `
...
 exports[`CdrFormError > using just the error prop (no slot) > renders correctly 1`] = `
 <div
-  class="cdr-form-error"
-  role="status"
+  class="cdr-form-error --active-error"
 >
-  <span
-    class="cdr-form-error__icon"
+  <div
+    class="cdr-form-error__icon cdr-form-error__icon--active"
```
cviterirei: Instead of adding `--active` you could use v-show.

src/components/formError/__tests__/__snapshots__/CdrFormError.spec.js.snap
saagude: Changed

cviterirei (approved):

mhewson (approved):

PR #235, opened 2026-03-03
V17 release

cviterirei:
## Description

...

### Design:

### Cross-browser testing:
- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] iOS
- [ ] Android

### Unit testing:

### A11y:
- [ ] Meets WCAG AA standards

### Documentation:


mhewson:
Hey Carlos, the only thing we need to get resolved is UMD build reliability because it looks like it can fail with error: too many open files.

Touchpoints: vite.umd.config.ts, rollupOptions.mjs
This is a release pipeline risk.

Maybe we can:

Inline the UMD config chain: The strongest signal is that the failure happens during config loading, not the library build. We could temporarily move the browserTargets and the shared UMD-specific rollup options directly into vite.umd.config.ts to kill the nested import chain and reduce file-handle pressure.

Swap rolldownOptions for rollupOptions: Since the failure stack is hitting the new Rolldown-backed path, we could fall back to standard rollupOptions for the UMD config if we don't strictly need Vite 8’s new behavior there yet.

Bump ulimit in CI: As a quick "unblock" for the release, we can raise the file descriptor limits in the runner, though I’d prefer one of the boring/stable config fixes above so developers don't keep hitting it locally.

What do you think? I’d suggest starting with inlining the config just to keep the release path as simple as possible.

mhewson (approved):

saagude (approved):

PR #238, opened 2026-04-23
Tokens direct imports

cviterirei:
## Summary
Updates Cedar to consume the aligned @rei/cdr-tokens package directly and removes the old dist-path shims. This branch also refreshes Cedar’s token-driven styles, stories, and type aliases so they match the new token layout and exported subpaths.

## Details
- Switches Cedar token imports to direct package subpaths:
  - `@rei/cdr-tokens/rei-dot-com/scss/cdr-tokens`
  - `@rei/cdr-tokens/rei-dot-com/css/cdr-tokens.css`
  - `@rei/cdr-tokens/docsite/json/...`
  - `@rei/cdr-tokens/rei-dot-com/types/...`
- Removes the previous `@rei/cdr-tokens/dist/...` fallback imports.
- Updates Storybook preview and generated template styles to use the new token entrypoints.
- Aligns Cedar palette overrides with the new token names and membership palette files.
- Updates token-backed enums and types in Cedar so stories and components use the regenerated token names.
- Refreshes surface-related story args and summaries to use token-backed options instead of hardcoded legacy values.
- Updates build config and lockfile to match the new token package dependency shape.
- Keeps Cedar building cleanly against the updated token package layout.

## Missing / Dependency
- This PR still depends on the matching rei-cedar-tokens work being available.
- It is reviewable as a stacked PR, but it is not standalone-mergeable until the tokens package is published and Cedar can drop the local workspace link.


saagude (approved):

.storybook/preview.ts
```diff
@@ -1,7 +1,7 @@
 import type { Preview } from '@storybook/vue3';
 
 // Import Cedar tokens CSS (provides CSS variables)
-import '@rei/cdr-tokens/dist/rei-dot-com/css/cdr-tokens.css';
+import '@rei/cdr-tokens/rei-dot-com/css/cdr-tokens.css';
```
mhewson: this should switch over from the deep link to the stable contract path 

import '@rei/cdr-tokens/css';

src/components/abstract/styles/CdrAbstract.module.scss
```diff
@@ -1,4 +1,4 @@
-@use '@rei/cdr-tokens/dist/rei-dot-com/scss/cdr-tokens' as tokens;
+@use '@rei/cdr-tokens/rei-dot-com/scss/cdr-tokens' as tokens;
```
mhewson: same here 
there is a consumer contract section on the tokens storybook that you should see all the available paths 
@use '@rei/cdr-tokens/scss';

src/types/other.ts
```diff
@@ -1,52 +1,197 @@
+import type { CdrColorBackgroundTokens } from '@rei/cdr-tokens/rei-dot-com/types/foundations/cdr-color-background';
```
mhewson: we should also update the module specific imports 

import type { CdrColorBackgroundTokens } from '@rei/cdr-tokens/types/cdr-color-background';
import type { CdrColorBorderTokens } from '@rei/cdr-tokens/types/cdr-color-border';
import type { CdrProminenceTokens } from '@rei/cdr-tokens/types/cdr-prominence';
import type { CdrRadiusTokens } from '@rei/cdr-tokens/types/cdr-radius';
import type { CdrSpaceTokens } from '@rei/cdr-tokens/types/cdr-space';

src/types/other.ts
```diff
@@ -1,52 +1,197 @@
...
-  | 'scale-6'
-  | 'scale-7'
-  | 'scale-8';
-export type SpaceScale = 'scale-0--1' | 'scale-3--4' | 'scale-3--5';
+
+/**
+ * Cedar surface options mapped to their canonical tokens. These maps are the
+ * migration source of truth for Storybook controls and component prop unions.
+ */
+export const surfaceBackgroundTokens = {
```
mhewson: This file is a stopgap from an earlier token architecture and should be deprecated. 
The intent is to move all token-based control surfaces (available backgrounds, radii, 
spacing options, etc.) to be dictated directly by the modularized token exports 
from @rei/cdr-tokens, not hardcoded lists in component repos.

The risk: these manual lists will drift from the actual tokens published by Cedar. 
For example, if design adds a new background color or removes deprecated spacing, 
this file won't auto-update—leaving Storybook and component props out of sync.

lets replace these exports with direct references to the published 
token modules. Instead of:
```ts
export const surfaceBackgroundTokens = { primary: '...', secondary: '...', ... }
export const surfaceBackgroundOptions = Object.keys(surfaceBackgroundTokens) as Background[];
```
Import the actual token keys from the Cedar contract and derive your options there:

```ts
import { CdrColorBackground } from '@rei/cdr-tokens';
export type Background = keyof typeof CdrColorBackground;
export const surfaceBackgroundOptions = Object.keys(CdrColorBackground) as Background[];
```

This way, design changes to tokens automatically propagate to your component controls.

src/utils/other.ts
```diff
@@ -1,5 +1,5 @@
 import * as tokens from '@rei/cdr-tokens';
-import { Breakpoint, Space } from '../types/other';
+import { Breakpoint, Space, spaceFixedOptions } from '../types/other';
```
mhewson: why do we have the two other.ts files? should these be one? 
This file has the same problems I would like to resolve as the other - being that its essentially recreating the token sets and forcing us to update allowed variants not only in tokens but then track them down here too - How much do you think we can migrate over to the updated tokens? lets update what we can and put a depreciation note in these files? 

I think we will probably have some items that dont belong in tokens - totally fine but maybe we can come up with a more descriptive name than other to inform us what these are really doing :P

mhewson (changes requested):

src/components/abstract/CdrAbstract.vue
```diff
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { useCssModule } from 'vue';
 import type { CdrAbstractProps } from './types';
-import type { Tag } from '../../types/other';
+import type { Tag } from '../../types/componentOptions';
```
mhewson, line 4: nice name

mhewson (approved):

